PR #1, opened 2019-05-15
Update requests and urllib3

withrocks:
* urrlib3 has security issues in 1.24.1
* Using the opportunity to fetch the latest version of requests

### DIFF
--- a/requirements-base.txt
+++ b/requirements-base.txt
@@ -53,7 +53,7 @@ rb>=1.7.0,<2.0.0
 redis-py-cluster==1.3.4
 redis>=2.10.3,<2.10.6
 requests-oauthlib==0.3.3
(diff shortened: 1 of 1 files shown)


requirements-base.txt
```diff
@@ -53,7 +53,7 @@ rb>=1.7.0,<2.0.0
 redis-py-cluster==1.3.4
 redis>=2.10.3,<2.10.6
 requests-oauthlib==0.3.3
-requests[security]>=2.20.0,<2.21.0
```
johandahlberg, line 56: Is there a reason that the version has been held back from 2.21.0 before? I'm just checking to make sure that it won't cause any issues to update it to 2.21.0 now?

withrocks: It should be OK, the tests are passing.

PR #3, opened 2019-05-20
Update documentation

withrocks:
Update the main documentation. This is all documenting how things should work in 1.0 (according to current plans).


docs/containers.md
```diff
@@ -0,0 +1,98 @@
+# Containers
+
+Samples can be added to a number of different containers, which can be extended by your plugins.
```
mariya, line 3: Maybe we should introduce the concept of "Item" early on and note that items refer to samples and sample-like objects?

docs/containers.md
```diff
@@ -0,0 +1,98 @@
...
+freezer.add(plate, row=1, col=1)
+```
+
+## Deeper nesting of containers
+
+You can nest containers deeper if you want to. Here for example, we define each shelve in the freezer so we can position each plate according to depth in the freezer too (3D):
+
+```
+freezer1 = Container("freezer1", rows=2, cols=1)
+shelve1 = Container("freezer1/shelve1", rows=2, cols=2)
```
mariya: the correct spelling is "shelf" for singular / "shelves" for plural

docs/high-level-design.md
```diff
@@ -0,0 +1,48 @@
...
+The core has already defined APIs for samples and containers. You would create a plugin that defines two things:
+
+* A simple Python script, written by you (research engineer, bioinformatician or developer) using the high level API. E.g.:
+
+	```
+	f = create_output_file("robotfile.txt")
+	for sample in input_samples():
+		f.write("{} {}".format(sample.name, sample.concentration))
+	```
+* A workflow (using a visual workflow designer) that describes the intended lifetime of the sample:
```
mariya: "lifetime" should be "lifecycle" i think. Also: "using a visual workflow designer" - shouldn't we be more explicit and say "Camunda", or do they have a say in swapping in their own workflow designer?

mariya (changes requested):
Really great job - this is mostly very clear! I'd like to polish these docs for general language issues, but I think we can take that much later, when we are closer to prime time.

johandahlberg:
This looks good. I've added some suggestions for updates as a PR here: https://github.com/commonlims/commonlims/pull/4 Its not automatically mergeable, but it should be simple enough to see my suggestions in https://github.com/commonlims/commonlims/pull/4/commits/600c6bc76341197852328bf9462e84a18239d5dc

docs/containers.md
withrocks: Right! And the verb is to shelve #ghoti.

docs/high-level-design.md
withrocks: Well, it's unclear at this point if they'll be able to do it in the browser or not and so on. Here I'm not telling them how exactly they'll do that. This is supposed to be a high-level design document. I certainly don't want to have them imagine that they need a different system for it - just "our visual workflow designer".

docs/containers.md
withrocks: I'm not entirely sure if the low level implementation detail, which the `Item` is will help them to understand the usage of the system. I think this might be clearer to start with. But I'm sure we'll have to revisit this!

withrocks:
Thanks for all your comments, great. I'll fix this accordingly and rebase.

withrocks:
@mariya Yeah, its excellent if we can stick to embarrassing-errors-only for now, since making this text production ready at this early stage would be wasting time. One must be able to understand it though!

We can always point back to this thread if people start blaming you for our English ;)

docs/high-level-design.md
withrocks: But yeah, it was lifecycle I was thinking of :)

PR #4, opened 2019-05-20
0 docs suggestions

johandahlberg:


docs/containers.md
```diff
@@ -14,6 +14,20 @@ Containers can have 2 or 3 dimensions, supporting e.g. tubes (2D), plates (2D) a
...
 When a sample is first created via the high-level API, it is not in a container. One can add it to one like this:
 
+TODO
+
+I think it could be helpful here to have an API that looks more like this:
+
+```
+a = Sample("A")
+plate = Container("c", rows=8, cols=12)
+plate[ContainerPosition(row="A", column="1")] = a
```
withrocks, line 24: Right. But this is only setting the dimensions of the `Container`. I.e. it's an 8x12 plate.

I had intentions for `plate["A1"]` (like we have in our extension framework) and similar to solve the case you're talking about.

docs/samples.md
```diff
@@ -83,3 +84,5 @@ a.volume -= 10.0
 It would be difficult to know what version to apply to each:
 
 ![Version inconsistency](./img/samples-04.png)
+
+Therefore we have decided that any time a sample is split a new sample entity must be created for the new aliqute. While, when a property is simply changed the sample identity remains the same.
```
withrocks, line 88: Cool, I'll remove the "we have decided" into a more "this is how it is" statement.

docs/samples.md
```diff
@@ -2,17 +2,18 @@ The core of any LIMS is managing samples and ensuring integrity of their data an
 
 # Auditing
 
+TODO: I think this section needs more work, in particular with regards to the discusions we have been having about versioning. /JD 2019-05-20
```
withrocks, line 5: Are there any particular inaccuracies we can fix now? All of this documentation will need several rounds.

withrocks:
Excellent feedback! I'll merge these in.

docs/high-level-design.md
```diff
@@ -4,38 +4,38 @@ Common LIMS is a framework to create custom LIMS systems in a consistent and sim
...
-		* Workflows
-		* Custom code
+    * Core
+    	* Addresses features you would expect from any information system
+    	* Related UI components
+    * Lab
+    	* Provides APIs for features that are common amongst most or all labs
+    	* Related UI components
+    * Plugins
+    	* Custom code
```
withrocks, line 14: Misses the "Workflows", perhaps unintentionally. But the workflows are defined in the plugins.

docs/samples.md
johandahlberg: For example the part "The history of samples is maintained by doing a full copy of them in the database before editing them", is something that I think is inaccurate now right? With the db design we looked at yesterday you don't need to copy the whole object on every write?

Also, I think that the part about GDPR is something that could perhaps go into a separate part of the documentation, that goes more into detail about audit tracking, data removal mechanisms, etc. But perhaps that is something that we can get back to revising later?

docs/samples.md
johandahlberg: :+1:

docs/containers.md
johandahlberg: I'm not sure I follow. The problem I see is that I see the designation `A1` as ambiguous. I.e. I find it difficult to know for sure if `A` or `1` represents the row?

docs/high-level-design.md
johandahlberg: Yeah. I though that referring to it here with out first having defined what it means could make the text more difficult to understand.

docs/containers.md
withrocks: Aha. The diff doesn't look like that's what being commented on, so it read like a misunderstanding.

But to comment on that suggestion then :) Sorry for how long it became, I hope it makes sense:

I believe it's much more common (if not a standard) that rows are named by letters and columns by numbers on plates. Googling "96 well plate" shows me only plates and diagrams using that convention, in cases where there are headers on it at all.

I think it should surely be an alternative to allow your suggestion (more on that below), but I personally wouldn't want to have to type that if the other is a convention as then it's more of a low level API than a high level one, and we should in my mind aim for a domain driven design and readability for the high level API. In our LIMS extension framework, we allow both.

It would also make even this document more hard to read for those that are accustomed to this convention, for `"A1" => "B1"` is so much easier to read than `ContainerPosition(row="B", column="1") => ContainerPosition(row="A", column="1")`.

While the latter is in some sense clearer, it's in the way if one is accustomed to the convention. It's similar to how we would never require one to say what are row and what are column indexes while using `i` and `j` to address locations in a matrix.

So, back to the implementation I had visioned. I was planning on an approach where the plate generally expects some kind of `Index` object, but also an "index-like" object. So given a string, it would try to parse it to an index object given the most common convention for that kind of container. Does that make sense?

docs/high-level-design.md
withrocks: Right, I guess we'll have to revisit this document several times. But I wonder if we shouldn't then restructure it a bit. 

It also depends on the audience reading this. When I wrote it I was thinking of devs, which would know what workflows are. It might make sense to read the documentation together at some later point and brainstorm on ideas.

docs/samples.md
withrocks: I totally agree with the GDPR thing, that's a typical example of what I think of as bad documentation, where one documents the thinking behind decisions rather than just how it works (keeping the decisions somewhere in an appendix, if anywhere). And yeah, missed that bit about the full copy, thanks for pointing out specific examples.

I'll fix this soon :)

withrocks: Another thing this kind of statement does is send people down an unnecessary rabbit hole they don't need to visit, e.g. "what's an append-only database? [15 mins on wikipedia later]. Ok where was I?".

docs/containers.md
johandahlberg: I think it makes sense if that is a strongly rooted convention in the domain to go with that, and I agree, that once you understand it `A1` is easier to read than `ContainerPosition(row='B', column='1')`. I like the "index-like" object idea. Thanks for taking the time to explain your reasoning. :+1:

docs/high-level-design.md
johandahlberg: I actually think that there are a wide variety of different understandings of what a workflow is, especially if you deal with people who have experience in bioinformatics. But I don't think that this is critical at this point, as you say we will need to get back to this later.

withrocks:
Sounds good :)

withrocks:
+1

PR #6, opened 2019-05-27
Add command to create example data

johandahlberg:
This adds the scaffolding for a command that can add test data to the db. This should be useful for developers who want to populate the system with data to start working on.


withrocks:
Looks good!

PR #11, opened 2019-05-29
Clims 66 - clean up samplecontainer

mariya:
This PR:

- cleans up sampleContainer.jsx
- adds tests for sampleContainer (still WIP)
- adds Sample helper class
- standardizes api usage for helper classes
- some naming improvements in other files


johandahlberg (approved):
I've looked this over, mostly to familiarize myself with the code. I cannot say that I understand this to the extent I can actually review neither the design or the implementation at this stage. I'll leave it to @withrocks to actually merge. :smile:

src/sentry/static/sentry/app/components/sampleTransitioner/sampleContainer.jsx
```diff
@@ -2,28 +2,13 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import SampleWell from './sampleWell';
 import { SampleLocation } from './sampleLocation';
+import { Sample } from './sample';
 
-export const SampleContainerType = {
+export const SampleContainerDirectionality = {
```
withrocks, line 7: I wonder if `SampleContainerSide` could be simpler to type and read? As in "it's on the source side (left), not the target side (right)".

src/sentry/static/sentry/app/components/sampleTransitioner/sampleContainer.jsx
```diff
@@ -34,203 +19,197 @@ export class SampleContainer extends React.Component {
...
-      // Get all samples with this row
-      const rowSamples = this.props.samples.filter(s => s.location.row === r);
-
-      key = `-1_${r}`;
-      cols.push(
-        <td key={key} style={this.getHeaderStyle(r, -1)}>
-          {this.getRowIndicator(r)}
-        </td>
+
+    onMouseOut();
```
withrocks, line 88: Is this recursive call accidental?

src/sentry/static/sentry/app/components/sampleTransitioner/sampleContainer.jsx
```diff
@@ -34,203 +19,197 @@ export class SampleContainer extends React.Component {
...
+  renderSampleWell(row, col, sampleId) {
+    const {
+      containerId,
+      onWellClicked,
+      onWellMouseOver,
+    } = this.props;
+
+    const key = `samplewell-${row}-${col}`;
+    const location = new SampleLocation(containerId, row, col);
+    const isTransitionSource = this.isTransitionSource(location);
```
withrocks: I wonder if some of these variables are unnecessarily set here, as they can be set in the return statement? Just thinking of readability since this indicates to my brain that this will be used in some way before returning.

withrocks (commented):
Looks great! I have a few questions, nothing that I think needs changing though except perhaps that call to `onMouseOut`.

src/sentry/static/sentry/app/components/sampleTransitioner/sampleContainer.jsx
mariya: It's not recursive. (A recursive call would be this.onMouseOut()). What this does is invoke the additional onMouseOut function that has been passed in as a prop.

src/sentry/static/sentry/app/components/sampleTransitioner/sampleContainer.jsx
mariya: Fixed this now.

PR #12, opened 2019-06-10
Fix JS tests

withrocks:
* Fixes or skips failing JS tests
* Upgrades Node


mariya:
This looks good to me. Merging!

PR #17, opened 2019-06-14
Clims 176 implement user task store redux

mariya:
This PR integrates Redux into the application and invokes an example action from the UserTask list (group.jsx). 

I've implemented actions to fetch the list of UserTasks from the API using Axios, and corresponding reducer. 

Nothing useful is done with data from the store in this particular PR, but you can see Redux logging the state if you navigate to the task list (http://localhost:8000/sentry/tasks/) and look in the console.

There are also tests for things.

If anyone wants a walkthrough, let me know!


johandahlberg (approved):
This looks good to me. If you could walk me through it when I'm back at the office this afternoon that would be great. I can merge this right a way if you want, though.

mariya:
Sounds good, let's go through this after lunch. Merge whenever you're ready!

PR #21, opened 2019-06-18
WIP: Fix linting and tests

withrocks:
WIP until #20 is merged.


setup.cfg
```diff
@@ -2,15 +2,15 @@
 python_files = test*.py
 # Note that the database is reused on tests, so run `pytest --create-db` before running the tests
 # if your models have changed
-addopts = --reuse-db --tb=native -p no:doctest -p no:warnings -s
```
johandahlberg, line 5: What does removing the `-s` do?

setup.cfg
```diff
@@ -2,15 +2,15 @@
...
+addopts = --reuse-db --tb=native -p no:doctest -p no:warnings
 norecursedirs = src/clims/legacy bin dist docs htmlcov script hooks node_modules .* {args}
 looponfailroots = src tests
 selenium_driver = chrome
 
 [flake8]
 ignore = F999,E501,E128,E124,E402,W503,E731,C901
 max-line-length = 100
-exclude = .venv/.git,*/migrations/*,node_modules/*,src/sentry/static/sentry/vendor/*,docs/*,examples/*
+exclude = .venv/.git,*/migrations/*,node_modules/*,src/sentry/static/sentry/vendor/*,docs/*,examples/*,src/clims/legacy,debug_toolbar
```
johandahlberg, line 13: Maybe later we'll want to auto-format all the legacy stuff to conform to the new standard? I don't mind this for now, though.

src/clims/api/endpoints/user_files.py
```diff
@@ -27,7 +20,7 @@ def get(self, request):
 
     def post(self, request):
         content = request.DATA["content"]
-        file_name = request.DATA["fileName"]
+        # file_name = request.DATA["fileName"]
```
johandahlberg, line 23: Perhaps add a comment here on why it is commented out rather than removed?

src/sentry/event_manager.py
```diff
@@ -664,17 +662,6 @@ def to_values(v):
             del data['stacktrace']
             # TODO(ja): Remove meta data of data['stacktrace'] here, too
 
-        # Exception mechanism needs SDK information to resolve proper names in
```
johandahlberg, line 667: Not quiet sure why this was removed.

johandahlberg (commented):
I guess all the stuff in `src/sentry/lang/**` was removed because it is nothing we need?

src/sentry/utils/db.py
```diff
@@ -43,20 +43,20 @@ def has_charts(db):
     return True
 
 
-def attach_foreignkey(objects, field, related=[], database=None):
+def attach_foreignkey(objects, field):
```
johandahlberg, line 46: Is it ok to change the function signature here?

johandahlberg (commented):
Any particular reason for removing `tests/sentry/models/test_organization.py`? I suppose that is a concept we will keep from Sentry?

setup.cfg
withrocks: Yeah no doubt. I was thinking too that the legacy stuff will be entirely revisited and autoformatted then.

src/sentry/utils/db.py
withrocks: Yes. This is only being used in one place and there were no parameters used there.

setup.cfg
withrocks: The `-s` flag was added by me by accident. It turns on a bunch of extra logging on success (turns off "capturing" logs).

src/sentry/utils/db.py
johandahlberg: :+1:

withrocks:
Doesn't look like `tests/sentry/models/test_organization.py` was removed, guess it was some github glitch?

src/sentry/event_manager.py
withrocks: I was removing that method `get_sdk_from_event` which is not used in our code.

src/clims/api/endpoints/user_files.py
withrocks: Should have done that :) But this will be revisited soon (TM).

PR #23, opened 2019-07-01
Clims 201 user task tag store

mariya:
This PR implements a Redux store for the tags that are used to filter User Tasks in the User Task List. It replaces the Reflux processTagStore.


johandahlberg:
lgtm

PR #26, opened 2019-08-07
No jira - document workflow concepts

mariya:


docs/workflows.md
```diff
@@ -0,0 +1,15 @@
+CLIMS uses the Camunda Workflow Engine under the hood, which in turn uses [BPMN conventions](https://camunda.com/bpmn/).
+
+Here we describe how core Camunda concepts have been adapted for use in CLIMS.
+
+__Process__ The entire workflow a sample might go through, for example, "DNA Sequencing" or "Reception QC". It is a directed acyclic graph comprised of activities or sub-processes.
```
withrocks: Minor point I guess: It's perhaps better to say "and" sub-processes, since I believe many think OR means XOR in text?

docs/workflows.md
```diff
@@ -0,0 +1,15 @@
...
+
+__Activity__ Also called a _step_, the smallest unit of work that a sample undergoes, for example, "Fragment Analyze".
```
withrocks: It's nice for us to use FA there for now since it directly relates to our work, but I'm not sure it makes much sense for many coming from another lab. A more generic example might be better, e.g. "Measure sample quality"?

docs/workflows.md
```diff
@@ -0,0 +1,15 @@
...
+
+__Process Instance__ A specific instance of a Process, as applied to a sample. When a sample is assigned to a process, an instance is created for that process and all its sub-processes and activities. The relationship between a sample and Process Instance is one-to-many.
```
withrocks: It's true that a process instance is created when a sample is assigned to a process, but not that instances will be created of sub-processes and activities. That happens only if the process reaches that point. Some activities and sub-processes will never be instantiated (e.g. if the work is cancelled or some path is taken causing that not to happen).

docs/workflows.md
```diff
@@ -0,0 +1,15 @@
...
+
+__Activity Instance__ A specific instance of an Activity, that has a status (e.g. "Ready", "Done"). An Activity Instance can be automated or require human intervention.
+
+__User Task__ An Activity Instance that requires human intervention.
```
withrocks: A "User Task" is an Activity that requires human intervention (not instance).

docs/workflows.md
```diff
@@ -0,0 +1,15 @@
...
+
+__Work Batch__ A bundle of User Tasks that is assigned to a user and has a status. This is a CLIMS-specific concept. When a batch of samples are assigned to a specific Process, one or more User Tasks are created for each of the samples. Usually these User Tasks are processed and resolved together by a specific user, e.g. when a lab technician runs "Reception QC" on an entire plate of samples. For this reason, it makes sense to bundle User Tasks together into Work Batches. If, for example, one or more samples do not pass QC, new UserTasks will be created for these deviating samples that are then bundled into a new Work Batch.
```
withrocks: Here it's also the case that the User Tasks are only created when the process reaches a "User Task" activity, i.e. it doesn't necessarily happen directly after assigning samples to a process.

docs/workflows.md
```diff
@@ -0,0 +1,15 @@
+CLIMS uses the Camunda Workflow Engine under the hood, which in turn uses [BPMN conventions](https://camunda.com/bpmn/).
```
withrocks: I wonder if "uses BPMN conventions" should instead be "implements the BPMN standard" and include a link to https://www.omg.org/spec/BPMN/2.0/ too.

withrocks (changes requested):
Looks cool! I've added a few comments/clarifications.

docs/workflows.md
mariya: Fixed

docs/workflows.md
mariya: Fixed

docs/workflows.md
mariya: Fixed

docs/workflows.md
mariya: Fixed

docs/workflows.md
mariya: Fixed

docs/workflows.md
mariya: Fixed

mariya:
Please take another look!

PR #29, opened 2019-08-12
Clims 208 rename user task to work batch

mariya:
I renamed UserTask to WorkBatch everywhere in the JS. Assumes that the endpoint /api/0/organizations/sentry/user-tasks has been renamed to /api/0/organizations/sentry/work-batches


withrocks (approved):

withrocks:
Looks good!

PR #32, opened 2019-08-15
Rename UserTask to WorkBatch

withrocks:
We're renaming the UserTask model to WorkBatch. This change was mostly automatic and I then ran migrations for the renamed models.


src/sentry/api/endpoints/project_tags.py
```diff
@@ -35,3 +35,31 @@ def get(self, request, project):
             )
 
         return Response(data)
+
+
+# TODO: Move to clims
+from sentry.api.bases.organization import OrganizationEndpoint
+
+
+class OrganizationWorkBatchPropertiesEndpoint(OrganizationEndpoint):
```
johandahlberg: Is this related to the rename?

withrocks: No, thanks fixed it!

PR #34, opened 2019-08-15
Add camunda setup

withrocks:
Script and config for setting up Camunda locally.

### DIFF
--- a/middleware/camunda/local-server.xml
+++ b/middleware/camunda/local-server.xml
@@ -1,0 +1,175 @@
+<?xml version='1.0' encoding='utf-8'?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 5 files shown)


middleware/camunda/local-server.xml
```diff
@@ -0,0 +1,175 @@
...
+    <Resource name="jdbc/ProcessEngine"
+              auth="Container"
+              type="javax.sql.DataSource"
+              factory="org.apache.tomcat.jdbc.pool.DataSourceFactory"
+              uniqueResourceName="process-engine"
+              driverClassName="org.postgresql.Driver"
+              url="jdbc:postgresql:clims"
+              defaultTransactionIsolation="READ_COMMITTED"
+              username="camunda"
+              password="changeit"
```
johandahlberg: Don't know, but perhaps we should generate a random password on each machine instead?

johandahlberg:
Looks good. Just one minor comment, as you can se above.

middleware/camunda/local-server.xml
withrocks: This is just for development purposes, as the name of the file implies (local). I wonder if we should wait for that until we create the docker images?

withrocks: I'll fix it :)

johandahlberg (approved):

PR #35, opened 2019-08-15
Add work batch and user task endpoints

withrocks:
This includes various changes mainly to provide an aggregate endpoint for querying for user tasks.

### DIFF
--- a/src/clims/api/urls.py
+++ b/src/clims/api/urls.py
@@ -25,6 +25,7 @@ from .endpoints.work_batch_notes_details import WorkBatchNotesDetailsEndpoint
 from .endpoints.plugin_actions import PluginActionsEndpoint
 from .endpoints.plugin_views import PluginViewsEndpoint
 
+from .endpoints.workflow import WorkflowEndpoint
 urlpatterns = patterns(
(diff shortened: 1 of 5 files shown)


src/clims/models/camunda.py
```diff
@@ -0,0 +1,43 @@
...
+
+# Models for camunda. None of these is created by CLIMS, but we require them here for querying them more easily
+# when the API isn't enough
+
+
+class CamundaProcess(models.Model):
+    __core__ = False
+
+    id = models.TextField(db_column='id_', primary_key=True)
+    rev = models.IntegerField(db_column='rev_', primary_key=True)
```
mariya: Should primary_key be true here?

src/clims/api/urls.py
```diff
@@ -34,7 +35,15 @@
         name='clims-api-0-sample-details'
         ),
 
-    # USER TASKS
+    # Workflow
+    url(r'^organizations/(?P<organization_slug>[^\/]+)/workflow/aggregate/task/$',
```
mariya, line 41: Move this to the earlier commit

mariya (changes requested):
See comments

withrocks:
I've fixed these issues now.

PR #41, opened 2019-08-23
Clims 54 task list ui

mariya:
Frontend for user task list.


src/sentry/static/sentry/app/views/taskList/index.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+class TaskList extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
+  componentDidMount() {
+    const {getTags} = this.props;
+    getTags();
```
withrocks, line 16: I wonder why the component does this?

src/sentry/static/sentry/app/views/taskList/tasks.jsx
```diff
@@ -0,0 +1,123 @@
...
+      body = <LoadingError message={errorMessage} onRetry={getTasks} />;
+    } else if (tasks.length > 0) {
+      body = this.renderProcesses();
+    } else {
+      body = this.renderEmpty();
+    }
+    return body;
+  }
+
+  groupTasksByProcess() {
```
withrocks, line 38: I wonder if this could be achieved in shorter code with more functional style (reducer), similar to what this person is doing: https://medium.com/@edisondevadoss/javascript-group-an-array-of-objects-by-key-afc85c35d07e

tests/js/spec/redux/actions/workBatch.spec.jsx
```diff
@@ -0,0 +1,106 @@
...
+  });
+
+  const mockWorkBatch = {
+    id: 4,
+    name: 'Test3',
+    organization: 1,
+    handler: 'somehandler2',
+    created: '2019-06-12T13:07:13.490Z',
+    extra_fields: '',
+    num_comments: 0,
```
withrocks, line 38: Ahh, we have python names for these fields. I'll have to fix that later :)

withrocks (approved):
Looks really good! I just have some questions, but I can merge this right away.

src/sentry/static/sentry/app/views/taskList/index.jsx
mariya: Well, eventually we will want filtering of user tasks by tags. We already have a redux action to fetch tags so that's just a step in that direction.

tests/js/spec/redux/actions/workBatch.spec.jsx
mariya: Ok :)

PR #42, opened 2019-08-28
Clims 54 tests for task list ui

mariya:
This PR:

- adds tests for User Task List
- documents how to test Redux-connected components
- improves User Task List code with more detailed PropTypes, constants for frontend magic numbers, cleanup of unnecessary styles, and other minor code tweaks

### DIFF
--- a/src/sentry/static/sentry/app/components/task/processListItem.jsx
+++ b/src/sentry/static/sentry/app/components/task/processListItem.jsx
@@ -11,8 +11,10 @@ class ProcessListItem extends React.Component {
     this.state = {
       showTasks: true,
     };
-    this.flexWidth = 300;
-    this.flexMargin = 2;
+    this.layout = {
+      flexWidth: 300,
+      flexMargin: 2,
+    };
   }
 
   toggleTasks() {
@@ -30,8 +32,7 @@ class ProcessListItem extends React.Component {
           name={name}
           count={count}
           taskDefinitionKey={taskDefinitionKey}
-          flexWidth={this.flexWidth}
-          flexMargin={this.flexMargin}
+          layout={this.layout}
           key={i}
         />
       );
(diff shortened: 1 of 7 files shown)


src/sentry/static/sentry/app/components/task/processListItem.jsx
```diff
@@ -65,7 +72,15 @@ ProcessListItem.propTypes = {
   processDefinitionKey: PropTypes.string.isRequired,
   processDefinitionName: PropTypes.string,
   count: PropTypes.number.isRequired,
-  tasks: PropTypes.arrayOf(PropTypes.shape({})),
+  tasks: PropTypes.arrayOf(
+    PropTypes.shape({
+      name: PropTypes.string.isRequired,
+      count: PropTypes.number.isRequired,
+      taskDefinitionKey: PropTypes.string.isRequired,
+      flexWidth: PropTypes.number,
```
withrocks: I personally prefer having `view` related parameters either outside of the business model or within a specifically named variable (e.g. `viewState`). Just for readability.

src/sentry/static/sentry/app/views/taskList/tasks.jsx
```diff
@@ -50,18 +47,19 @@ class Tasks extends React.Component {
...
+            tasks: [],
+            count: 0,
+            processDefinitionKey,
+            processDefinitionName,
+          };
+      r[processDefinitionKey].count += count;
+      r[processDefinitionKey].tasks.push(prunedTask);
+
+      return r;
+    }, {});
```
withrocks, line 62: I was wondering if this could become like just a one-liner with a helper function or library that would in turn use `reduce` (I guess). Similar to what this person is doing: https://stackoverflow.com/a/34890276/282024

Both approaches seem to contain a lot of boilerplate to my javascript-untrained eyes.

withrocks:
Looks great! I have two minor discussion points only. Let me know if I should merge directly

src/sentry/static/sentry/app/views/taskList/tasks.jsx
mariya: Not really, that would work if I were creating a simple object of the format 'processkey': [tasks]. But I'm actually creating a nested object like 'processkey': {tasks: [tasks], count: x ...}

src/sentry/static/sentry/app/components/task/processListItem.jsx
mariya: Fixed!

PR #43, opened 2019-09-09
Update documentation on samples

withrocks:


withrocks:
The build is failing for unrelated reasons (timeout in acceptance and some issue with the pluggy plugin in pytest). Since this is just a doc change I think it should still go through and we'll look into this issue later.

GitEdvard:
This is straightforward to me!

PR #44, opened 2019-09-10
Upgrade django and rest framework

withrocks:


withrocks:
This is WIP at the moment because I'm fixing some tests that used to work here but aren't now.

The only way to code review this is commit-by-commit.

src/sentry/web/urls.py
```diff
@@ -65,13 +65,13 @@ def init_all_applications():
...
-    from django.db.models import get_apps, get_models
+    # TODO: Is this still needed in Django 1.9+
+    from django.apps import apps
 
-    for app in get_apps():
-        try:
-            get_models(app)
-        except Exception:
-            continue
+    try:
```
johandahlberg, line 71: Not sure I understand why you need to do this?

src/sentry/plugins/base/manager.py
```diff
@@ -46,40 +42,7 @@ def all(self, version=1):
             yield plugin
 
     def _register_work_batches(self, class_path):
-        # TODO: Not caching this load for now for simplicity, fix that in the child class
```
johandahlberg, line 49: Was this purposefully removed here? Just checking, since there was not mention of making changes to the plugin manager in the commit message.

src/sentry/api/serializers/rest_framework/sentry_app.py
```diff
@@ -17,8 +17,14 @@ def validate(self, data):
             if scope not in valid_scopes:
                 raise ValidationError(u'{} not a valid scope'.format(scope))
 
+    def to_internal_value(self, data):
```
johandahlberg, line 20: Shouldn't this wrap some internal class? Or am I misunderstanding these docs: https://www.django-rest-framework.org/api-guide/fields/#custom-fields

src/sentry/conf/server.py
```diff
@@ -265,7 +265,7 @@ def env(key, default='', type=None):
     'django.contrib.sites',
     'django.contrib.staticfiles',
     'crispy_forms',
-    'debug_toolbar',
```
johandahlberg, line 268: Maybe we should add some `TODO` comment here, remarking why this was removed?

src/sentry/conf/server.py
```diff
@@ -133,6 +133,8 @@ def env(key, default='', type=None):
     if url.scheme == 'mysql':
         DATABASES['default']['ENGINE'] = 'django.db.backends.mysql'
 
+ALLOWED_HOSTS = ['localhost']
```
johandahlberg: Should `localhost` always be allowed? Or is it enough to have that enabled in debug mode? https://docs.djangoproject.com/en/2.2/ref/settings/#allowed-hosts

johandahlberg (approved):
This looks very good. I had a few minor comments, but other than that I think this looks good. Thanks for the well formatted git-log. That made this much easier to follow that it would otherwise have been.

One question of a more general nature; in cfea03a7d5caa30b4557fe990ca34629b0367526 all of the imports are moved around. Looking at the release notes for Django 1.9, I can see nothing there that would indicate that this behavior has changed in the framework. Could you walk me through why this is necessary now?

src/sentry/web/urls.py
withrocks: Yeah, I'm not sure that it was needed in this version. They say that they're doing this to "force import" (in the method comment above) and I'm just changing the function call to get models. But I've added a TODO to investigate if this is needed or not.

src/sentry/plugins/base/manager.py
withrocks: Good catch - should have written something about this. It's OK that this is removed - it will be addressed in future commits. I assume (some time since I did this) that I got an issue while pushing in the patch and decided to resolve it by removing this legacy code.

src/sentry/conf/server.py
withrocks: Will fix.

src/sentry/conf/server.py
withrocks: Excellent catch. According to the docs localhost should be added if we're running in DEBUG mode. I noticed that when running `lims devserver` the configuration is run before running the setup of environment variables, so this DEBUG flag wasn't set correctly. So I fixed that.

However, then I also noticed that even if we do set it correctly, localhost isn't being added for some reason as it should be (also in 1.9). Rather than digging too deep into that at the moment, I'm just adding it when we are in DEBUG mode: https://github.com/commonlims/commonlims/pull/44/commits/4dbcbc7ba2efcd7404a7c5538462eaea2c77a394

src/sentry/conf/server.py
withrocks: Fixed!

withrocks: ... that is, I've added `TODO: removed temporarily` as a `git blame` (or `git log --follow`) describes the reason.

src/sentry/api/serializers/rest_framework/sentry_app.py
withrocks: Good question. It took me some time to wrap my head around this reading the docs as well.

The thing is that the methods are used to map between primitive and non-primitive data types (where dictionaries and lists are also considered primitive in this case).

`The to_internal_value() method is called to restore a primitive datatype into its internal python representation.`

In the case of `api_scopes`, the primitive data that crosses the wires will be a list of strings. And then when we use it on the server side, we'll still just need a list of strings.

One might then ask why we implement `to_internal_value` at all. That's because the framework requires that for `WritabelField`s:

`Note that the WritableField class that was present in version 2.x no longer exists. You should subclass Field and override to_internal_value() if the field supports data input.` (from the doc you've linked to).

johandahlberg: Thank you for the explanation! :bowing_man:

src/sentry/conf/server.py
johandahlberg: Sounds good.

PR #45, opened 2019-09-13
Sample list - preliminary chagnes

withrocks:
These are some rather minor changes that need to be in place before a larger patch with support for the new `Substance` type.


src/clims/models/plugin_registration.py
```diff
@@ -0,0 +1,31 @@
...
+from sentry.db.models import (Model, sane_repr)
+
+
+class PluginRegistration(Model):
+    """
+    Lists every registration of a plugin, ensuring that configuration created by a plugin can be traced to a particular
+    version of that plugin.
+
+    Take for example a look at the ItemType model. When a plugin registers a particular
+
```
GitEdvard: Here, the text stops in the middle of a sentence!

src/clims/models/plugin_registration.py
```diff
@@ -0,0 +1,31 @@
...
+    If you start the service with a particular version of a plugin (e.g. clims_genetics@1.0.0) and then update
+    it only at 2.0.0, we'll only have a record of those two plugins on the system, even if there are more
+    available on PyPi.
+    """
```
GitEdvard, line 12: I dont catch the point here.

GitEdvard (commented):
Ok, I have comments on the comments. Tell me when you want to merge. 

Is substance going to replace ActivityInstance? Did I got that right?

src/clims/models/plugin_registration.py
withrocks: Thanks! That happens once in a while for me.

src/clims/models/plugin_registration.py
withrocks: I think it's too much detail there anyways. I'll remove it. Also removed that cut sentence.

PR #46, opened 2019-09-16
Add extensible types and substances

withrocks:
The generic model `Substance` replaces the `Sample`. It's built on the `ExtensibleType` which allows users to easily add new fields to them.


withrocks:
I suggest reading the documentation and the tests first in order to understand what this PR tries to achieve.

docs/deep-dive/substances.md
```diff
@@ -0,0 +1,88 @@
...
+When the system registers your plugin, it finds your `GemstoneSample`. Based on it, the system creates an entry in the table `ExtensibleType`, which registers your `GemstoneSample` class. It also creates two entries in the `ExtensiblePropertyType` table, for each of the properties.
+
+Now you can create an instance of the `Substance` like this:
+
+```
+s1 = GemstoneSample()
+s1.color = "red"
+s1.comment = "some comment"
+s1.save()
+```
```
GitEdvard, line 37: I think it would be helpful if you showed how this instantiation is represented in each of the four db tables, like...

docs/deep-dive/substances.md
```diff
@@ -0,0 +1,88 @@
...
+
```
GitEdvard, line 38: ExtensibleType:

docs/deep-dive/substances.md
```diff
@@ -0,0 +1,88 @@
...
+    dict(name='comment', raw_type=ExtensiblePropertyType.STRING, display_name='Comment'),
+]
+
+substance_type = substances.register_type('GemstoneSample', 'substances', plugin, properties=properties)
+```
+
+We will now have one `ExtensibleType` entry in the DB as well as two `ExtensiblePropertyType` entries.
+
+The next step is to create the `Substance`. One can do this via the `substances` singleton like this:
+
```
GitEdvard, line 61: I think it's more clear with ... create the 'substance' instance.

docs/deep-dive/substances.md
GitEdvard: |  id  | name |
|  1  | 'Gemstone' |

GitEdvard: Then Extensible Property Type

GitEdvard: | Id  | ext type id | name |
| 1   | 1               |  'color' |

src/clims/models/substance.py
```diff
@@ -0,0 +1,84 @@
...
+    A substance is a model that describe primarily samples and aliquots but they
+    can also represent other elements that can be combined with a sample or aliquot to make
+    another substance.
+
+    Substances can have plugin-defined key/values which are defined in `ExtensibleProperty` instances.
+    All of these extra properties have a detailed version history that can be directly retrieved through
+    the API.
+
+    NOTE: Use the substances service class (`from clims.services import substances`) to create or update substances and properties
+    as otherwise business rules will be broken, e.g. you might update a substance but not its properties.
```
GitEdvard, line 18: I dont like when the text goes outside the border :)

src/clims/models/extensible.py
```diff
@@ -0,0 +1,184 @@
...
+    a django migration. They are defined programmatically in the plugin and registered
+    when running `lims upgrade`.
+
+    These types are examples of those that will be extensible, meaning that they can
+    be customized by plugins. Currently the Substance is the only one that's implemented.
+
+    * Substance
+    * Container
+    * Project
+    """
```
GitEdvard, line 20: I'm concerned by the potential performance problems that may come with a large number of samples in a system. From that perspective, is it a good idea to have separated tables for container, project and substances? The important thing here is in this case that ExtensiblePropertyValue are split up according to these base types. This would protect e.g. project handling to be slowed down because of many samples in the system.

GitEdvard: I also want to mention that there might be a need to list all 'things' in a system in one single list, i.e. substances, containers and projects. I think of the barcodes in Chiasma and Clarity. Every object has a globally unique barcode.

src/clims/models/extensible.py
```diff
@@ -0,0 +1,184 @@
...
+        app_label = 'clims'
+        db_table = 'clims_extensiblepropertyvalue'
+
+
+class ExtensibleProperty(Model):
+    """
+    A class for properties that can be hooked up to `Extensible` objects.
+
+    Examples:
+
```
GitEdvard: There is no examples here, skip.

src/clims/models/extensible.py
```diff
@@ -0,0 +1,184 @@
...
+            ret.float_value = value
+        elif raw_type == ExtensiblePropertyType.STRING:
+            ret.string_value = value
+        elif raw_type == ExtensiblePropertyType.BOOL:
+            ret.bool_value = value
+        elif raw_type == ExtensiblePropertyType.JSON:
+            ret.string_value = json.dumps(value)
+        else:
+            raise ValueError("Unexpected raw type {}".format(raw_type))
+        ret.save()
```
GitEdvard: There might be a need for validating user input before saving to db. So, maybe it might be a good idea to defer saving data to db?

tests/clims/models/test_substance.py
```diff
@@ -0,0 +1,197 @@
...
+            extensible_type=self.gemstone_sample_type,
+            organization=self.org,
+            properties=props)
+
+        assert substance.version == 1
+        assert {prop.version for prop in substance.properties.all()} == {1}
+
+        # NOTE: The high level link between substances and properties (where they are versioned together)
+        # is (currently) only supported via the SubstanceService, so it's not possible to change properties on the object
+        # we just created above.
```
GitEdvard, line 79: Shorter lines please?

tests/clims/models/test_substance.py
```diff
@@ -0,0 +1,197 @@
...
+    def test_updating_properties_via_update_or_create_updates_version(self):
+        props = dict(preciousness='*o*', color='red')
+
+        substance = substances.create(
+            name='gem-sample-001',
+            extensible_type=self.gemstone_sample_type,
+            organization=self.org,
+            properties=props)
+
+        assert substance.version == 1
```
GitEdvard, line 74: I dont really see the connection between substance version and substance property version in these tests. I assume substance version will increment each time substance.save is called. But what if, for example substance.save is called with the underlying substance not changed?

docs/deep-dive/substances.md
```diff
@@ -0,0 +1,88 @@
...
+    comment = TextField()
+```
+
+The GemstoneSample object is in fact a proxy for a range of database models:
+
+    * ExtensibleType
+    * ExtensiblePropertyType
+    * Substance
+    * ExtensibleProperty
+
```
GitEdvard, line 27: In code, there is ExtensiblePropertyValue as well. I think it would be nice with a description of how ExtensiblePropertyValue and ExtensibleProperty relates to each other, and also how version of a property comes in. I think it would be really helpful to have this explanation to lean back on when one want to refresh the knowledge of how it works.

docs/deep-dive/substances.md
```diff
@@ -0,0 +1,88 @@
...
+)
+```
+
+This will create one `Substance` where:
+
+    * It is marked as version 1
+    * It has the properties `comment` and `color` set
+    * It has been type checked against the substance type we created before
+    * It has no origins or parents as it's not derived from any sample
+
```
GitEdvard, line 82: I miss an overview showing 1) version of substance, 2) version of property, 3) depth och substance, 4) parent of substance, 5) origin of substance.

GitEdvard: This document is close related to the sample.md document. Maybe we should put them more close together? Rename it to substance overview, then substance deep dive?

src/clims/models/extensible.py
GitEdvard: Shouldnt saving to db be done in the substance service?

GitEdvard (commented):
Hello!
As you can see, I have quite a few comments on the doc section. I'm a bit hesitant about the docs, it might be a waste to have too extensive documentation, so you have to use your own judgement here. On the other hand, I really think that I would be helped with a good doc of how substances work under the hood.

GitEdvard:
I also have more general considerations about the work on commonlims. In my view, the first priority should be to have a shippable version as soon as possible. And that would imply a very slimmed down and lightweight first version. 

At first, I thought that the extensible types felt a bit complicated, and I feared gold plating. When looking at it now, I realize that probably it will be very easy to use in code, which is good. But I'm concerned by future performance problems. I think most user cases will be about handling the latest version of samples, not previous samples. So maybe it's an idea to have two tables, one containing all latest values, which is faster, and another containing all historical values, which would be slower. 

Once again, this might be a concern that are to be handled when the problem arises? However, shuffling production data between tables might fell a bit uncomfortable.

GitEdvard:
I have another question.... In my pool-ticket (clims-222) it's mentioned that substance should be a rest endpoint. I think that you have forgotten that?

src/clims/models/extensible.py
withrocks: I understand your concern and think it's great you're bringing it up early.

Note that substance, container and project will all have their own tables, because they'll have different requirements. We'll for example need ancestry information for `substances` but not `projects`. It's only the properties tables that will be reused for all of these. 

But this would still mean we'll have a huge properties table. That does not worry me because when we do queries in a DB we usually (preferably always for larger tables) really don't query the table under the hood, that's just a convenient abstraction. We are rather querying an index into the table. So in fact there will be different indexes for the `project` to `properties` relation than are for the `subject` to `properties` table. This way of approaching the problem moves most of the performance concerns from "domain developer" hands to the hands of people that know everything about DB peformance, in this case the postgres core team.

It also means that there will be a lot of extra data created (the indexes), but that will be an constant increase so it should not be a concern. (Also, data is cheap.)

I'm certain that with a moderately beefy DB server, this won't be an issue, even for millions of entries, but we'll write performance tests to show that later. Then we'll also address the lack of indexes.

Finally I'd like to add that I think we should never optimize prematurely, as long as there is a straightforward way to do it later, usually achieved by separation of concerns. If it would turn out I'm wrong, we would be able to fix things by only changing the data layer.

src/clims/models/extensible.py
```diff
@@ -0,0 +1,184 @@
...
+
+    TYPE_CHOICES = (
+        (STRING, 'string'),
+        (JSON, 'json'),
+        (INT, 'int'),
+        (FLOAT, 'float'),
+        (BOOL, 'bool'),
+    )
+
+    extensible_type = models.ForeignKey('clims.ExtensibleType', related_name='property_types', null=True)  # NOMERGE: Remove null
```
withrocks: Just noticed this now, fixing.

docs/deep-dive/substances.md
withrocks: Good point! I'll fix this.

docs/deep-dive/substances.md
withrocks: Can do.

docs/deep-dive/substances.md
withrocks: Yep, will fix.

src/clims/models/extensible.py
withrocks: Good catch with not saving here! I'll fix that, it works in this case because the caller has a transaction, but that might not always be the case.

Regarding validation, the REST layer should certainly validate the input there. Not sure the service layer should, that it should rather trust that callers do that (since they will be in-process callers, like plugins or other code). Still uncertain about that aspect, we should definitely have more discussion on that.

src/clims/models/extensible.py
withrocks: Yes!

src/clims/models/substance.py
withrocks: I'll fix this! This should have been automatically fixed by the linter (we have a limit at 100). I'll look into why not.

docs/deep-dive/substances.md
withrocks: Yes, I'll address the first point. I think I'll do another runthrough later to address the second point.

src/clims/models/extensible.py
GitEdvard: Ok, thanks for the explanation, and interesting with the index thing. I also agree with the last thing, never to optimize prematurely.

GitEdvard: I want to mention the reason why I have concerns about performance. 1) clarity is very slow. I dont know why, udf system might contribute to this. 2) A general uncomfortness because I have never worked with this kind of solution before. I just have a diffuse feeling that many linked large tables may be slow. It would defenitely have some value if we could have a solution that is stress-tested in a production environment before.

docs/deep-dive/substances.md
withrocks: I've added `./deep-dive/extensible.md` that goes into detail. Check out my rebased commit.

docs/deep-dive/substances.md
withrocks: In the `extensible.md` doc I just added I do this.

docs/deep-dive/substances.md
withrocks: After giving it some more thought, I think I'll address these in a later commit as it's still somewhat a moving target :)

src/clims/models/extensible.py
withrocks: Yes. I think your concern is very valid, not in the least given our past experience with that system.

But as mentioned, I think it's improper implementation that causes this in most or all cases. Databases are extremely well optimized for this sort of thing and have been gradually improving since the 70s, but they are often misused. I look forward to doing some load tests on this. (I'd also like to mention that load testing on this without change would not show excellent performance since the indexes aren't all in place right now - I'm deferring that job to when the API isn't as much of a moving target.)

withrocks:
Thanks for your excellent feedback. Regarding your general comments:

I decided to push the REST api changes to another PR so it would be easier to review this. It's coming later today!

You have a good point that perhaps most of the data will be required most of the time. I totally agree with that, but I think also that can be optimized later. A typical solution for that is to regularly (e.g. every month) do a archival for historical data to another table.

But here again indexes should come to the rescue. The reason is that we have a field that indicates if a property is `latest` or not. If we create an index that includes that field for the most common queries, we effectively have created two tables (the indexes are two) one for queries for the current data and one for the non-current.

tests/clims/models/test_substance.py
withrocks: Good catch! Something seems to have been lost from an earlier version of the code. I will fix this.

withrocks: ... it was just the test that needed to be updated. I'm now also asserting that the substance also got an increased version.

But yes, the substance should get a new version as well as the properties that have been updated.

At the moment, I'm assuming that the name will not be updated - and if a plugin wants to allow users to update the name they'll have to have a versionable name property that can be changed(!) I consider this a potential weakness which I'd like us to look more into later. But as it stands, only the properties are updated.

tests/clims/models/test_substance.py
withrocks: Yes, I'll remember that :) Was sure the autofixer was taking care of it :/

PR #47, opened 2019-09-18
Substance endpoints

withrocks:
* An initial sketch for the endpoints for substances
* Formatting issues in pre-commit

### DIFF
--- a/src/clims/api/serializers/models/substance.py
+++ b/src/clims/api/serializers/models/substance.py
@@ -1,0 +1,67 @@
+from __future__ import absolute_import
+
+from clims.models.substance import Substance
+from rest_framework import serializers
+from rest_framework.fields import Field
+
+
+class PropertyValueField(Field):
+    def to_representation(self, obj):
+        return obj
+
+    def to_internal_value(self, data):
+        return data
+
+
+class PropertyTypeSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+
+
+class PropertyOverviewSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    type = PropertyTypeSerializer()
+    value = PropertyValueField()
+
+
+class PropertiesField(Field):
+    def to_representation(self, obj):
+        ret = dict()
+        for prop in obj.filter(latest=True):
+            prop_type = prop.extensible_property_type
+            display_name = prop_type.display_name if prop_type.display_name else prop_type.name
+            ret[prop_type.name] = dict(value=prop.value, display_name=display_name)
+        return ret
+
+    def to_internal_value(self, data):
+        return data
+
+
+class VersionField(Field):
+    def to_representation(self, obj):
+        return obj
+
+
+class ExtensibleTypeField(Field):
+    def to_representation(self, obj):
+        return "{}.substances.{}".format(obj.plugin.name, obj.name)
+
+    def to_internal_value(self, data):
+        return data
+
+
+class SubstanceSerializer(serializers.Serializer):
+
+    id = serializers.IntegerField(read_only=True)
+    version = serializers.IntegerField(read_only=True)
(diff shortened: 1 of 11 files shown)


src/clims/services/substance.py
```diff
@@ -25,32 +25,40 @@ class SubstanceService(object):
...
         if not new_name:
             new_name = "{}:{}".format(parent.name, uuid4())
 
-        child = Substance(name=new_name, organization=parent.organization, extensible_type=parent.extensible_type)
+        child = Substance(
+            name=new_name,
+            organization=parent.organization,
+            extensible_type=parent.extensible_type)
         child.depth = parent.depth + 1
         child.save()
```
GitEdvard, line 57: I have yet another thought about this. In this setup, there is no distinction between substance factory and substance repository. So you cant create an instance in memory only, without also creating it in db. This might have implications in future. In any way, it's probably best to deal with that potential problem when it comes. It also seems to me that this is the setup for django itself?

src/clims/api/serializers/models/substance.py
```diff
@@ -0,0 +1,66 @@
...
+class ExtensibleTypeField(Field):
+    def to_representation(self, obj):
+        return "{}.substances.{}".format(obj.plugin.name, obj.name)
+
+    def to_internal_value(self, data):
+        return data
+
+
+class SubstanceSerializer(serializers.Serializer):
+
```
GitEdvard, line 53: I'm not sure I understand these serializers fully. How is it with the foreign key navigation properties that exists within substance, like project, and the many-to-many navigation properties, origins and parents are not here, but properties are. Shouldn't depth be here?

tests/clims/api/endpoints/test_substances.py
```diff
@@ -0,0 +1,60 @@
...
+class SubstancesTest(APITestCase):
+    def test_get_substances(self):
+        substance = testutils.create_substance('sample-001', {})
+
+        url = reverse('clims-api-0-substances', args=(substance.organization.name,))
+        self.login_as(self.user)
+        response = self.client.get(url)
+
+        assert response.status_code == 200, response.content
+        assert len(response.data) == 1
```
GitEdvard: I may have misunderstood this. But if another test creates a sample with with testutils.create_substance(), isn't this test going to fail then?

GitEdvard:
Ok, now I have reviewed. Tell me when you want to merge!

src/clims/services/substance.py
withrocks: In this setup, I would argue that django has the repository layer. One can create objects via that layer without saving them. However, when calling the service, I believe that it should actually save the objects, because it's supposed to be used by high level code (the REST API and the plugin API).

src/clims/api/serializers/models/substance.py
withrocks: In the serialization "layer" one can decide what the REST API users will see. One argument is that they should see everything, and that makes a lot of sense in that it's simple, but it also gives access to data that the frontend might not need.

My thinking was to defer that decision to later and only start with what we know the frontend requires. But I would also be fine with us just serializing everything directly - so please add what you need too :)

The rest_framework has serializers that serialize whole models and make this easier, but in this case I wrote our own because the substance relates to all those different models to do what it needs to do so I assumed it would be simpler to just throw together a serializer.

GitEdvard: Ok, thats clear to me now.

src/clims/services/substance.py
GitEdvard: Ok!

tests/clims/api/endpoints/test_substances.py
withrocks: Yes nice catch. It seems that these API test cases aren't transactional. I'll change this to just requiring one more than the previous run.

PR #48, opened 2019-09-23
Sample submission files

withrocks:


src/clims/services/substance.py
```diff
@@ -167,4 +167,64 @@ def register_type(self, name, org, plugin, category='default', properties=None):
...
+    def __get__(self, obj, type=None):
+        return obj.properties[self.prop_name]
+
+    def __set__(self, obj, value):
+        # TODO: Validate
+        self.validate(obj, value)
+        obj.properties[self.prop_name] = value
+
+
+class SubstanceBase(object):
```
GitEdvard: Hm, it's a little bit misleading that an entity class is in a service module. But, it's good that classes referring to each other  are in vicinity of each other.

src/sentry/runner/initializer.py
```diff
@@ -97,6 +94,27 @@ def init_plugin(plugin):
...
+    try:
+        plugin_version = plugin.version
+    except AttributeError:
+        plugin_version = "NA"
+
+    try:
+        registered_name = "{}.{}".format(plugin_type.__module__, plugin_type.__name__)
+        # NOTE: Registration currently happens for all organizations:
+        for org in Organization.objects.all():
+            PluginRegistration.objects.get_or_create(
```
GitEdvard: I think will need a walkthrough, in future,  of how plugins are to be used, and how they relates to organizations. Here all found plugins are registered to all organizations. Is that the intention?

devboot-js
```diff
@@ -0,0 +1,32 @@
...
+    echo "Installing Python 2.7.15..."
+    pyenv install 2.7.15
+fi
+
+if [ ! -d ~/.pyenv/versions/clims2 ]; then
+    echo "Creating virtualenv ..."
+    pyenv virtualenv 2.7.15 clims2
+fi
+
+pyenv activate clims2
```
GitEdvard: this file is called from devboot-conda. Is this really going to work with only conda installed?

setup.py
```diff
@@ -50,10 +51,20 @@
...
+    m = GIT_REGEX.search(line)
+    if m:
+        return m.group(1)
+    return line
+
 
 def get_requirements(env):
     with open(u'requirements-{}.txt'.format(env)) as fp:
-        return [x.strip() for x in fp.read().split('\n') if not x.startswith('#')]
+        return [parse_req(x.strip()) for x in fp.read().split('\n') if not x.startswith('#')]
```
GitEdvard, line 67: Is this handling reqs of both types?  1) django-debug-toolbar>=1.9.1,<1.10.0, and 2)  git+git@github.com:commonlims/django-templatetag-sugar.git@master#egg=django_templatetag_sugar?

GitEdvard:
Nice work, tell when you want to merge!

src/clims/services/substance.py
withrocks: Yes I agree, this is temporary :) I just don't want them in the models, as it's not a django model and we have circular dependency issues when loading them.

src/sentry/runner/initializer.py
withrocks: Right, for now they are just registered for all organizations, with the possibility to build on top of this in the future. It would probably be preferable not to have the same plugins registered for all plugins, but I think we can live with it like this in 1.0.

devboot-js
withrocks: Good catch, seems I didn't clean up properly!

setup.py
withrocks: Yes it will - kind of. This is an ugly temporary hack to support git requirements without changing too much. So the setup.py file will just know that it depends on `django-templatetag-sugar`, while it will be installed in the makefile with information on from where to get it (since it's not pypi).

withrocks:
Thanks for the feedback! I have minor fixes that I'll push in before merging.

src/clims/handlers.py
```diff
@@ -9,6 +9,16 @@ class MultipleHandlersNotAllowed(Exception):
     pass
 
 
+class Context(object):
```
GitEdvard: (sidenote) In future, I think it would be nice with a more descriptive class name. Context is super generic name.

src/clims/services/extensible.py
```diff
@@ -0,0 +1,54 @@
...
+            plugin=plugin)
+
+        existing_property_types = {item.name: item for item in extensible_type.property_types.all()}
+
+        for prop in properties:
+            prop_type = existing_property_types.get(prop['name'], None)
+            if not prop_type:
+                prop_type = ExtensiblePropertyType(extensible_type=extensible_type)
+            for key, val in prop.items():
+                setattr(prop_type, key, val)
```
GitEdvard, line 49: I have a questionmark here. I thought that properties so far could be of any of these types: int, float, bool, string. But here, one single property can have a list of properties, with dedicated names, of its own?

src/clims/services/extensible.py
```diff
@@ -0,0 +1,54 @@
+from __future__ import absolute_import
+
+from clims.services.substance import ExtensibleBaseField
+from clims.models import ExtensibleType, ExtensiblePropertyType
+from django.db import transaction
+
+
+class ExtensibleService(object):
```
GitEdvard, line 8: This service has to do with configuring an extensible, not handling extensible instances. Maybe that should be indicated by its name?

src/clims/services/extensible.py
```diff
@@ -0,0 +1,54 @@
...
+    def register(self, plugin, extensible_base):
```
GitEdvard, line 9: (sidenote) When trying to understand what this code does, it was a very good help to understand that it was the ExtensibleType and ExtensiblePropertyType tables that are updated. I could then relate to the docs about the extensibles. Mayby that should be indicated somehow in method names?

src/clims/services/extensible.py
```diff
@@ -0,0 +1,54 @@
...
+                raw_type=field.raw_type,
+                display_name=field.display_name)
+            properties.append(prop_type)
+
+        name = "{}.{}".format(extensible_base.__module__, extensible_base.__name__)
+        extensible_type = self._register_model(name, 'default', plugin, properties=properties)
+        return extensible_type
+
+    @transaction.atomic
+    def _register_model(self, name, org, plugin, category='default', properties=None):
```
GitEdvard: I wonder if this method might be a little bit too generic. If the extensible type doesn't exist before, its straightforward. But, if the extensible type already exist, I had to look into details to understand that new properties can be added, but properties may not be altered. One alternative would be, method 1: register new model, method 2: add property to extensible.

tests/clims/models/test_substance.py
```diff
@@ -211,6 +211,29 @@ def do_asserts(child):
...
 
 
+class TestSubstanceType(TestCase):
+    def setUp(self):
+        pass
+
+    def test_it(self):
+        # TODO: This is really an extensible test
+        plugin = testutils.create_plugin()
+        extensibles.register(plugin, GemstoneSample)
```
GitEdvard, line 221: It would be nice with some documentation on how plugins are to be used in commonlims. (maybe there already are?). It seems to me that one single database may contain multiple plugins, and each extensible type is associated with a certain plugin? How does this relate to e.g. stockholm and snpseq working with the same lims?

withrocks: Yes. There is some documentation already and I think we should certainly document it more when we have less of a moving target.

The idea is that one can have 0..n plugins running and even different plugins depending on the organization. The types are as you say associated with a plugin.

So we will be able to have e.g. one sample object as required by lab1 and another by lab2. They can be as much alike the other or differ as much. They can also inherit from another if required. In the end they are just key/value pairs in the `Extensible` system.

src/clims/services/extensible.py
withrocks: Yes, this is still not fully laid out. I think this method could just update the properties when they need to be altered. I was thinking that one of us fixes it when the use case is required (and certainly before 1.0)

src/clims/services/extensible.py
withrocks: I'm aware that this is a lot to take in, since we have the whole model story and then the higher-level abstractions on top. I hope that we can name them more clearly with time.

src/clims/services/extensible.py
withrocks: At the moment it only handles that, but I was planning to move some of the stuff that handles instances to it too later (they are currently in substance.py, which doesn't make that much sense).

src/clims/handlers.py
withrocks: I agree, `HandlerContext` would probably be better.

withrocks: Fixed!

src/clims/services/extensible.py
withrocks: I think this is just confusing naming. The properties are actually supposed to be property_types.

I've fixed it now!

PR #49, opened 2019-10-03
238, Import excel files in generic file handling

GitEdvard:
Purpose:
Prepare for import sample submission files in plugins. Here Excel files may be imported in StructuredFileMixin. 

Comments1:
The tests for gemstone sample submission is temporarily placed in commonlims, although they belong to commonlims-snpseq. Please review it anyway. 

Comments2:
I added new versions of gemstone sample submission example files. I was not sure if the old ones already were used. Maybe we can replace them instead?

Comments3:
I never got this to work in the ui. I think something in the ui code is buggy, and I stopped care about it. However, the code in test, _create_organization_file, parts of it is taken from /commonlims/src/clims/api/endpoints/substance_file.py.


src/clims/models/file.py
```diff
@@ -16,13 +17,21 @@ def as_excel(self):
...
+            workbook = load_workbook(tmp.name, data_only=True)
+
+        return workbook
 
     def as_xml(self):
         raise NotImplementedError()
 
     def as_csv(self):
-        from clims.services import Csv
+        # TODO: Add @lru_cache to the method when ready
```
johandahlberg, line 33: I'm note sure I see the purpose of setting a `lru_cache` here. What was your reasoning?

johandahlberg (approved):
Looks really nice. I had a small question. But I will go ahead and merge this, and then we can take that in another PR if necessary.

johandahlberg:
And we can have a look at the UI code later. :+1:

src/clims/models/file.py
GitEdvard: It's something that remained from a pair session with Steinar. In beginning as_csv was decorated with @lru_cache, and it caused problems. Then we moved it to a TODO :)

PR #50, opened 2019-10-03
Make the dev setup easier

johandahlberg:
This introduces new instructions for settting up the dev-env, including how to setup password login for the clims database. The default user for the database is `clims` now, and it uses a randomly generated password to log in.

It also fixes some things in the camunda db-install script to make it a bit more robust.


withrocks:
Looks good!

PR #51, opened 2019-10-04
WIP: Substance refactoring

withrocks:
This is a large refactoring of the substance API and related fixes.


src/clims/api/endpoints/substance.py
```diff
@@ -5,34 +5,35 @@
...
     def get(self, request, organization):
         # TODO: Add version as query parameter
-        # Prefetch the latest properties on the object
-        prefetch = Prefetch(
-            'properties',
-            queryset=ExtensibleProperty.objects.filter(latest=True))
-
-        queryset = Substance.objects.all().prefetch_related(prefetch)
+        # TODO: Get the queryset through the service
+        queryset = SubstanceVersion.objects.filter(latest=True).prefetch_related('properties')
```
withrocks: I'll change this real quick.

src/clims/models/extensible.py
```diff
@@ -1,10 +1,52 @@
...
+
+    latest = models.BooleanField(default=True)
+
+    properties = models.ManyToManyField('clims.ExtensibleProperty')
+
+    class Meta:
+        abstract = True
+
+
+class SubstanceVersion(ExtensibleVersion):
```
withrocks: I'll move this!

src/clims/models/extensible.py
```diff
@@ -125,18 +136,44 @@ class ExtensibleProperty(Model):
...
-        self.is_versioned = kwargs.get('is_versioned', True)
+
+    def _get_field_from_type(self):
+        raw_type = self.extensible_property_type.raw_type
+        if raw_type == ExtensiblePropertyType.INT:
+            return "int_value"
+        elif raw_type == ExtensiblePropertyType.FLOAT:
+            return "float_value"
+        elif raw_type == ExtensiblePropertyType.STRING or raw_type == ExtensiblePropertyType.JSON:
+            # TODO: We need to parse json
```
withrocks: Will remove this todo

src/clims/api/serializers/models/substance.py
```diff
@@ -43,19 +39,18 @@ def to_representation(self, obj):
...
 
 class SubstanceSerializer(serializers.Serializer):
-
-    id = serializers.IntegerField(read_only=True)
+    id = serializers.CharField(read_only=True)
     version = serializers.IntegerField(read_only=True)
     name = serializers.CharField()
     properties = PropertiesField()
-    extensible_type = ExtensibleTypeField()
+    type_full_name = ExtensibleTypeField()
```
johandahlberg: Perhaps a `CharField` here?

src/clims/models/extensible.py
```diff
@@ -1,10 +1,44 @@
...
-from sentry.db.models import (Model, sane_repr)
+from sentry.db.models import (Model, sane_repr, FlexibleForeignKey)
 import json
 
 
+class ExtensibleModel(Model):
+    """
+    Base class for extensible models
+    """
+    __core__ = True
```
johandahlberg: Should `__core__` be set to `True` in the base class, or should it be `False`, so that any plugins extending it defaults to that?

src/clims/models/extensible.py
```diff
@@ -1,10 +1,44 @@
...
+    """
+    Represents a particular version of an Extensible
+
+    Parent/Child relationship is always between a particular version of an Extensible.
+    """
+    __core__ = True
+
+    # This field is only populated if a user renames the extensible. In this case we record
+    # the name change here in the version model. This makes handling unique constraints simpler.
+    previous_name = models.TextField(null=True)
```
johandahlberg, line 30: I'm not sure I understand what renaming here implies? Rename the class?

src/clims/models/extensible.py
```diff
@@ -125,18 +128,43 @@ class ExtensibleProperty(Model):
...
 
-    version = models.IntegerField(null=False, default=1)
-
-    # True if this is the latest version of this instance
-    latest = models.BooleanField(null=False, default=True)
-
     extensible_property_type = models.ForeignKey("clims.ExtensiblePropertyType")
 
-    extensible_property_value = models.ForeignKey("clims.ExtensiblePropertyValue")
+    # Supported values (TODO: add more)
```
johandahlberg, line 133: Nice to more this into here, instead of having `ExtensiblePropertyValue`.

src/clims/models/extensible.py
```diff
@@ -125,18 +128,43 @@ class ExtensibleProperty(Model):
...
+        elif raw_type == ExtensiblePropertyType.FLOAT:
+            return "float_value"
+        elif raw_type == ExtensiblePropertyType.STRING or raw_type == ExtensiblePropertyType.JSON:
+            return "string_value"
+        elif raw_type == ExtensiblePropertyType.BOOL:
+            return "bool_value"
+        else:
+            raise AssertionError("Unexpected raw type {}".format(raw_type))
+
+    def _serialize(self, value):
```
johandahlberg, line 155: Do I understand correctly here that any json data is stored as string values? Perhaps we should add a todo on add the postgress specific [JSONField](https://docs.djangoproject.com/en/2.2/ref/contrib/postgres/fields/#django.contrib.postgres.fields.JSONField) here?

src/clims/models/substance.py
```diff
@@ -17,39 +18,38 @@ class Substance(Model):
...
     """
     __core__ = True
 
     def __init__(self, *args, **kwargs):
         super(Substance, self).__init__(*args, **kwargs)
 
-    name = models.TextField()
-
-    version = models.IntegerField(default=1)
+    name = models.TextField(null=True)  # TODO: null=False
```
johandahlberg: Why is `null=True` ok now?

src/clims/services/application.py
```diff
@@ -0,0 +1,14 @@
+from __future__ import absolute_import
+
+from clims.services.extensible import ExtensibleService
+from clims.services.substance import SubstanceService
+
+
+class ApplicationService(object):
```
johandahlberg, line 7: Perhaps `CLimsApplication` or just `Appliction`, would be a more user friendly name for this? To signal to the API user that this is the full application context?

src/sentry/testutils/cases.py
```diff
@@ -377,6 +387,34 @@ def assertWriteQueries(self, queries, debug=False, *args, **kwargs):
         with context:
             func(*args, **kwargs)
 
+    def create_organization(self):
```
withrocks: Turns out there already was a fixture for this (in ./fixtures). I'll remove this.

src/clims/services/extensible.py
```diff
@@ -1,11 +1,56 @@
...
+        self._implementation_cache.clear()
+        self._model_cache.clear()
+
+    def create(self, name, klass, organization, properties=None):
+        """
+        Creates an extensible and saves it to the backend.
+        """
+        if isinstance(klass, six.string_types):
+            klass = self.get_implementation(klass)
+        instance = klass(name=name, organization=organization, _app=self._app)
```
johandahlberg, line 28: One question that struck me as I read this is, will plugins be per organization, or for the entire installation?

src/clims/services/extensible.py
```diff
@@ -1,11 +1,56 @@
...
 
+class ExtensibleTypeNotRegistered(Exception):
+    pass
+
+
 class ExtensibleService(object):
+    def __init__(self, app):
+        self._app = app
+        self._implementation_cache = dict()
+        self._model_cache = dict()
```
johandahlberg, line 17: Is the `model_cache` really necessary? Or does it just add unnecessary complexity?

johandahlberg: Also, I'm kind of wondering if we use the implementation_cache for? To me it seems it is not really quite a cache (which to me would imply that it can be cleared at any point without consequence), instead it just seems like a list of implementations. Perhaps we should rename it to implementations instead? And not clear it on the `clear_cache` method?

src/sentry/api/base.py
```diff
@@ -63,6 +63,13 @@ class Endpoint(APIView):
     parser_classes = (JSONParser, )
     permission_classes = (NoPermission, )
 
+    @property
+    def app(self):
+        if not hasattr(self, "_app"):
```
johandahlberg, line 68: Cool pattern. Haven't seen it before, but it looks neat.

src/sentry/models/counter.py
```diff
@@ -103,32 +104,37 @@ def create_counter_function(sender, using, **kwargs):
...
-          new_val int;
-        begin
-          loop
-            update sentry_projectcounter set value = value + delta
-             where project_id = project
-               returning value into new_val;
-            if found then
-              return new_val;
-            end if;
+    try:
```
johandahlberg, line 107: I don't quite understand the functions of this `Counter` code - I guess this is fine, but looking through the sentry code base it only appears to be used once (outside of the tests), so perhaps it is something we could get rid of?

src/sentry/plugins/base/manager.py
```diff
@@ -9,12 +9,14 @@
```
johandahlberg, line 9: I'm not going to go over this very closely right now, because I think we need to do a full review/overhaul of this later anyways. Does that sound ok?

src/sentry/testutils/cases.py
```diff
@@ -377,6 +387,29 @@ def assertWriteQueries(self, queries, debug=False, *args, **kwargs):
...
+            name='tests_utils.create_plugin', version='1.0.0', organization=org)
+        return plugin
+
+    def register_extensible(self, klass, plugin=None):
+        plugin = plugin or self.create_plugin()
+        ret = self.app.extensibles.register(plugin, klass)
+        return ret
+
+    def create_substance(self, klass, name=None, **kwargs):
+        # TODO: This shouldn't be necessary as there should be only one app instance per test
```
johandahlberg: Isn't this fixed now?

tests/clims/api/endpoints/test_substances.py
```diff
@@ -1,57 +1,76 @@
...
         response = self.client.get(url)
         len_before = len(response.data)
 
-        substance = testutils.create_substance()
+        second = self.create_gemstone()
         response = self.client.get(url)
 
         assert response.status_code == 200, response.content
+
+        # NOTE: This may not work if tests run async on the same DB
```
johandahlberg: Am I wrong in assuming that this works as expected now?

johandahlberg (approved):
Looks cool. I had some questions, and comments, look through them and see if there is anything you want to change, and then we can merge this.

src/clims/api/serializers/models/substance.py
withrocks: Yes!

src/clims/models/extensible.py
withrocks: Good question. The `__core__` flag is only being used to decide if a table should be exported or not. It makes sense that this data should not be exported as it's not settings data. I'll set it to `False`. I've also created a JIRA ticket to go through this everywhere `CLIMS-243` (next milestone).

withrocks: I would then say it makes sense to have the type info core, even if it's defined in a plugin. Of course I would prefer that we would set this data only via plugins, but it may be good to be able to export everything that is a "setting".

src/clims/models/extensible.py
withrocks: Ahh. I was thinking of changing this to name and populate it always with the current name. It will be clearer and this approach was used as I was probably in a micro optimizing mode/daze. We can look into that later.

But this is when you want to rename the instance. So the idea is that the `Extensible` model always has the current name of the thing in question (e.g. `sample-1`), but the `ExtensibleVersion` has the whole history (either using this approach or the aforementioned approach of always saving the current name).

src/clims/models/extensible.py
withrocks: Yes :) That was the original (and natural) approach. The only reason I used the `*Value` class was that I had made a bad premature optimization decision (to handle versioning), which lead to that rather unnatural solution.

src/clims/models/extensible.py
withrocks: Yes, you're right! I agree. I'll add a ticket.

src/clims/models/substance.py
withrocks: It's not :) This should have been marked with # NOMERGE (which I do when I want to fix it before I send the PR).

It can sometimes be convenient to allow nulls while dealing with the django migrations. This is not an obvious case of that. Assume some strange stuff was happening.

withrocks: ... so I'm fixing this.

src/clims/services/application.py
withrocks: The plugin API user would always get an instance of this which is just named `app` for now. They would never have to instantiate this class.

src/clims/services/extensible.py
withrocks: It's absolutely necessary. Otherwise we would have to go the DB to fetch type metadata everytime one instantiates an extensible.

So an operation that everyone would think would be free, e.g. `sample = Sample()` would lead to just about the largest overhead possible, a call to the DB. (Unless I'm missing something?)

withrocks: Good catch about not clearing the implementation cache in clear cache. That was a mistake which I think I did blindly because of the naming :) I'll rename it to implementations.

johandahlberg: I think you are right about the query going to the database. Originally I was thinking that this might be cached by the ORM layer, but I read up on that now, and that does not seem to be the case. I'm mostly asking about it since I think that it looks a little bit like a premature optimization (and caches are often difficult to reason about). However, I good with keeping it this way. :+1:

withrocks: Of course knowing what's premature and what's not is an art, but I would say that a good rule of thumb is never to go to the DB without a reason. A better example of premature opt in our context would be more on the level of "should I go to the hard disk on the server or just to memory".

The downside of not optimizing obvious stuff right away is that then you miss out on months of testing of that feature. My view is that if it should be there and it doesn't take much time to add it, it should go in right away (getting-things-done-style).

src/clims/services/extensible.py
withrocks: They are bound to each organization, but at the moment we register them for all organizations.

The reason for that is that I think we need to discuss implementation details for how to, since things are auto registered, we should bind them to only organization. One idea is to not allow auto registration if there is more than one org in the system.

src/sentry/api/base.py
withrocks: :) I think it could be criticized also for being rather unclear, that the `_app` isn't set in the constructor. I thought about it and decided to throw it in like this for now - we can easily fix it later.

src/sentry/models/counter.py
withrocks: Yes. I'd like to get rid of it. They use it I believe to count the projects on the DB end for performance reasons.

I believe they use it since they have a huge instance in the cloud and probably an enormous amount of projects. We could easily just count the projects on the server side when required. I've added a TODO to fix it (below).

withrocks: ... in general I believe we should not use any DB functions in the end product, as having the business logic only in the server makes it much easier to test.

src/sentry/plugins/base/manager.py
withrocks: Yes, sounds good!

src/sentry/testutils/cases.py
withrocks: Yeah, don't really understand what I'm going on about there...

tests/clims/api/endpoints/test_substances.py
withrocks: I think my point here was that if we start running the tests async, the delta would possibly be more than 1. But if we use the django asynch features, the system creates one test database per process, so that should not be an issue. I'll remove the comment.

src/clims/services/extensible.py
johandahlberg: :+1:

PR #55, opened 2019-10-11
Fixing various stuff

johandahlberg:
This PR provides various fixes:
 - Setup of the db and test dbs in a more secure manner
 - Adding graphviz as a dependency
 - Make the script to create some test data use the new APIs
 - Fixes a bug I found when trying to create an API token
 - Re-enables a bunch of tests that were disabled, but that do seem to work  - this part I think would do well with some careful code review from @withrocks to see I've not been to heavy handed here.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 autopep8>=1.3.5,<1.4.0
-Babel
+Babel>=2.7.0,<3.0.0
 
 # Dependency of flake8
 # https://github.com/jaraco/configparser/issues/27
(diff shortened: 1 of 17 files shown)


requirements-dev.txt
```diff
@@ -7,6 +7,7 @@ Babel
 configparser!=3.5.2,!=3.5.3,!=3.7.0
 django-debug-toolbar>=1.9.1,<1.10.0
 flake8>=3.5.0,<3.6.0
+graphviz
```
withrocks: It wouldn't hurt to bind this to a minor version. Same goes for inflect and ipython which I think I added at some point :|

src/clims/runner/commands/create_example_data.py
```diff
@@ -11,10 +12,32 @@ def createexampledata():
...
+    from sentry.models.organization import Organization
 
-    # Import models here because they need to be wrapped with a Django context to work.
-    # /JD 2019-05-27
-    from clims.models.work_batch import WorkBatch
-    WorkBatch.objects.create(name="Test1", organization_id=1, handler="somehandler")
-    WorkBatch.objects.create(name="Test2", organization_id=1, handler="somehandler2")
-    WorkBatch.objects.create(name="Test3", organization_id=1, handler="somehandler2")
+    app = ApplicationService()
+    ioc.set_application(app)
```
withrocks, line 21: Note to self: We should allow that `ioc` to return a default version of the application if no one has called `set_application`. Then this kind of setup would be required in e.g. this case.

withrocks:
Looks great!

PR #56, opened 2019-10-14
Clims 245 - More info from substance REST API

johandahlberg:
This PR does to things:
 - It makes sure that properties get set when doing something like `Sample(prop1=1)`, which was not the case before
 - Each substance entry from the `/substances` endpoint will look something like
```
    {
        "id": 500,
        "name": "sample-360782",
        "properties": {
            "cool": {
                "display_name": "cool",
                "id": "297",
                "name": "cool",
                "value": 76.0
            },
            "erudite": {
                "display_name": "erudite",
                "id": "295",
                "name": "erudite",
                "value": 2.0
            },
            "moxy": {
                "display_name": "moxy",
                "id": "296",
                "name": "moxy",
                "value": 55.0
            }
        },
        "type_full_name": "clims.services.substance.SubstanceBase",
        "version": 1
    }
```


src/clims/api/serializers/models/extensible_property.py
```diff
@@ -0,0 +1,15 @@
...
+
+from rest_framework import serializers
+
+
+class ExtensiblePropertySerializer(serializers.Serializer):
+    id = serializers.CharField(read_only=True)
+    version = serializers.IntegerField(read_only=True)
+    name = serializers.CharField(read_only=True)
+    display_name = serializers.CharField(read_only=True)
+    value = serializers.SerializerMethodField(method_name="get_value_with_correct_type")
```
withrocks, line 12: I wonder what this does?

src/clims/api/serializers/models/substance.py
```diff
@@ -1,47 +1,17 @@
...
-class VersionField(Field):
-    def to_representation(self, obj):
-        return obj
+from rest_framework import serializers
+from rest_framework.fields import DictField
 
 
 class SubstanceSerializer(serializers.Serializer):
-    id = serializers.CharField(read_only=True)
+    id = serializers.IntegerField(read_only=True)
```
withrocks, line 11: This was an interesting thing. I decided to use a string, as it's an implementation detail that it actually is an integer in the backend. Since the API user will never have to treat the ID as anything but a string (they are for example not going to add two IDs), I felt that it should be a string. At the moment however, it really doesn't matter and we will probably never want to use anything else anyways (e.g. GUIDs). So perhaps it's right to just KISS it.

tests/clims/api/serializers/models/test_extensible_property.py
```diff
@@ -0,0 +1,51 @@
...
+from clims.models.extensible import ExtensibleProperty, ExtensiblePropertyType, ExtensibleType
+from clims.models.plugin_registration import PluginRegistration
+
+from sentry.models.organization import Organization
+
+
+class ExtensiblePropertyTypeSerializerTestCase(TestCase):
+
+    def setUp(self):
+        org, _ = Organization.objects.get_or_create(name="testorg")
```
withrocks: FYI: There is an organization available as a fixture for all tests in `self.organization`. I was using this pattern earlier but have moved to using only the fixtures.

tests/clims/models/test_substance.py
```diff
@@ -15,6 +17,42 @@ def register_gemstone_type(self):
...
+        class ExampleSample(SubstanceBase):
+            moxy = FloatField("moxy")
+            cool = FloatField("cool")
+            erudite = FloatField("erudite")
+
+        moxy = random.randint(1, 100)
+        cool = random.randint(1, 100)
+        erudite = random.randint(1, 100)
+
+        self.app.extensibles.register(example_plugin, ExampleSample)
```
withrocks: FYI: It's also possible to write `self.register_extensible` directly. Then you don't have to create the org or plugin.

withrocks (approved):
Looks good!

tests/clims/models/test_substance.py
johandahlberg: Neat. I will go back and use these fixtures.

PR #57, opened 2019-10-14
Create a view for substances

withrocks:
This PR includes several smaller commits that could benefit from being reviewed in isolation.


src/sentry/static/sentry/app/views/substances/substances.jsx
```diff
@@ -0,0 +1,95 @@
...
+    } else if (this.props.errorMessage) {
+      return <LoadingError />;
+    }
+
+    return (
+      <div>
+        <ReactTable
+          data={this.props.substances}
+          columns={[
+            {
```
withrocks: These columns will come from the API and will not be in the UI like this. Just want to get a preview out to trunk as soon as possible.

src/sentry/static/sentry/app/views/substances/substances.jsx
```diff
@@ -0,0 +1,95 @@
...
+              Header: 'Priority',
+              id: 'priority',
+              accessor: d => d.priority,
+            },
+            {
+              Header: 'Waiting',
+              id: 'days_waiting',
+              accessor: d => d.days_waiting,
+              Cell: row => (
+                <span>
```
withrocks: This should be a react component if we decide to use it, making the code more readable - again, just getting the preview out directly.

tests/js/spec/redux/actions/substance.spec.jsx
```diff
@@ -0,0 +1,106 @@
...
+describe('substance redux actions', function() {
+  beforeEach(function() {
+    moxios.install();
+  });
+
+  afterEach(function() {
+    moxios.uninstall();
+  });
+
+  const mockSubstance = {
```
withrocks, line 31: This test is a copy of the one in workbatches. I'll update this before removing WIP.

src/sentry/static/sentry/app/views/substances/substances.jsx
johandahlberg: :+1: Pephaps add a TODO in the code to indicate this_

src/sentry/static/sentry/app/redux/actions/substance.js
```diff
@@ -0,0 +1,102 @@
...
+
+export const substanceGet = id => dispatch => {
+  dispatch(substanceGetRequest(id));
+  return axios
+    .get(`/api/0/substances/${id}/`)
+    .then(res => dispatch(substanceGetSuccess(res.data)))
+    .catch(err => dispatch(substanceGetFailure(err)));
+};
+
+export const substanceGetIfNeeded = id => {
```
johandahlberg, line 82: I notice that this function is not used. Is it something that you plan to use later? Otherwise perhaps we should remove it?

src/sentry/static/sentry/app/redux/reducers/substance.js
```diff
@@ -0,0 +1,87 @@
...
+        ...state,
+        errorMessage: null,
+        loading: false,
+        byIds,
+      };
+    }
+    case SUBSTANCE_GET_FAILURE: {
+      return {
+        ...state,
+        errorMessage: 'Funky!',
```
johandahlberg: Perhaps we want a more informative error message here later? :wink:

src/sentry/static/sentry/app/redux/actions/substance.js
```diff
@@ -0,0 +1,102 @@
+import axios from 'axios';
+
+export const SUBSTANCES_GET_REQUEST = 'SUBSTANCES_GET_REQUEST';
+export const SUBSTANCES_GET_SUCCESS = 'SUBSTANCES_GET_SUCCESS';
+export const SUBSTANCES_GET_FAILURE = 'SUBSTANCES_GET_FAILURE';
+export const SUBSTANCES_TOGGLE_SELECT_ALL = 'SUBSTANCES_TOGGLE_SELECT_ALL';
```
johandahlberg, line 6: Do we also need an action to deselect all?

johandahlberg (commented):
I have started looking at this, but I still have some reviewing left to do. I'll also readily admit that the js stuff is not my specialty. To my eyes it looks alright - but take that with a grain of salt. Hopefully I'll find time to get back to this tomorrow.

src/sentry/static/sentry/app/redux/reducers/substance.js
withrocks: Nice catch :P

src/sentry/static/sentry/app/views/substances/substances.jsx
withrocks: Will add that!

src/sentry/static/sentry/app/redux/actions/substance.js
withrocks: We'll need this when we have the details view. Added it now as this was based on work batches, which had it.

src/sentry/static/sentry/app/redux/actions/substance.js
withrocks: This action handles both, as there is a flag in the action called `doSelect` which specifies if it should be toggled off or on.

src/clims/services/application.py
```diff
@@ -12,14 +13,21 @@ class ApplicationService(object):
     def __init__(self):
         self.extensibles = ExtensibleService(self)
         self.substances = SubstanceService(self)
+        self.containers = ContainerService(self)
 
 
 class InversionOfControl(object):
```
johandahlberg, line 19: Perhaps it would make sense to name this to something less generic? I don't know exactly what that would be right now, though.

src/clims/services/container.py
```diff
@@ -0,0 +1,50 @@
...
+
+from clims.services.extensible import ExtensibleBase
+from clims.models import Container, ContainerVersion
+
+
+class ContainerService(object):
+    def __init__(self, app):
+        self._app = app
+
+    # TODO: Some code-copying form SubstanceService
```
johandahlberg, line 11: Yeah. Lets get back to this when we do the project model/service.

src/clims/services/container.py
```diff
@@ -0,0 +1,50 @@
...
+class ContainerBase(ExtensibleBase):
+    """
+    A base object for defining custom containers.
+
+    Details:
+
+    Under the hood, this object wraps a Container object and its related Extensible* classes.
+    """
+
+    WrappedArchetype = Container
```
johandahlberg, line 45: Neat.

src/clims/api/endpoints/substance.py
```diff
@@ -24,7 +24,17 @@ def handle_results(qs):
             # the wrapper object directly.
             for entry in qs:
                 wrapper = self.app.substances.to_wrapper(entry)
-                ret.append(SubstanceSerializer(wrapper).data)
+                json = SubstanceSerializer(wrapper).data
+                json['position'] = {'index': "A:1", 'container': {'name': "cont1"}}
```
johandahlberg, line 28: I'd prefer moving this to the `create_example_data.py` script, so that we start building a "library" of example data that makes sense and that can easily be inserted into the system.

johandahlberg (approved):
I think a lot in the Substance/Extensible confusion has been cleared up here. Great work! I had some minor comments, but I'm good to merge when you are.

src/clims/services/application.py
withrocks: Perhaps you're right. This service should handle all containers so the name clearly describes what it's supposed to handle at least. We can always revisit this (next six months anyways).

src/clims/api/endpoints/substance.py
withrocks: Ahh, I think there is some misunderstanding. This is just pushing attributes on the entry that don't exist (not generating data), just so we can see the thing in the view. I'll remove it in the next commit to follow.

withrocks:
Cool! You can go ahead and merge then :)

src/clims/api/endpoints/substance.py
johandahlberg: Yeah. I understand that - my suggestion was to add it in the data instead. But as you say, we can get to that later.

PR #58, opened 2019-10-16
Fix url patterns in clims

johandahlberg:
The first argument to the `patterns` function is a base prefix for all routes in that list that follows. Since that was missing the first route was missed.


withrocks:
Nice that you solved it :)

withrocks (approved):

PR #60, opened 2019-10-17
Add support for downloading a demo sample file

withrocks:


src/clims/api/endpoints/substance_file.py
```diff
@@ -8,16 +8,34 @@
...
     permission_classes = (IsAuthenticated, )
 
     def get(self, request, organization):
-        # TODO: and write test
-        return Response({})
+        """
+        List all substance submission files in the system
+        `````````````````````````````````````````````````
+
+        :pparam string organization_slug: the slug of the organization the
```
johandahlberg, line 26: Typo?

src/clims/api/endpoints/substance_file_details.py
```diff
@@ -0,0 +1,74 @@
...
+        )
+        return response
+
+    def download_from_model(self, model):
+        file = model.file
+        fp = file.getfile()
+        size = file.size
+        name = model.name
+        content_type = 'application/octet-stream'
+        # TODO: file.headers is an empty list now, not a dictionary(?) We should use the header
```
johandahlberg, line 38: I don't really understand this comment. IT looks like you are using and `octet-stream` in the response? No?

src/clims/api/endpoints/substance_file_details.py
```diff
@@ -0,0 +1,74 @@
...
+        :auth: required
+        """
+
+        try:
+            submission_file = self.app.substances.get_submission_file(file_id)
+        except NotFound:
+            raise ResourceDoesNotExist
+
+        download_requested = request.GET.get('download') is not None
+        if download_requested:  # and (request.access.has_scope('project:write')) TODO
```
johandahlberg: I might be wrong, but isn't the `permission_classes` used to handle permissions?

src/clims/services/extensible.py
```diff
@@ -41,12 +41,16 @@ def get_implementation(self, extensible_type_full_name):
 
     def get_extensible_type(self, org, name):
         """
-        Returns an extensible_type by name. Can be cached.
+        Returns an extensible_type by name. May be from a cache.
```
johandahlberg: May be -> Maybe?

src/clims/services/substance.py
```diff
@@ -73,13 +78,30 @@ def all_qs(self):
...
+    def get_submission_file(self, file_id):
+        try:
+            return OrganizationFile.objects.get(id=file_id)
+        except OrganizationFile.DoesNotExist:
+            raise NotFound("Can't find file with id '{}'".format(file_id))
+
     @transaction.atomic
-    def load_file(self, organization, full_path, fileobj):
+    def load_file(self, organization, full_path, fileobj, add_timestamp=False):
+        # TODO: Decide what we want to happen here. Should the user be able to load files
```
johandahlberg, line 94: I agree. The file name shouldn't matter. The file content is what should be checked and validated to not add e.g. duplicate samples.

src/sentry/static/sentry/app/views/substances/uploadSubstancesButton.jsx
```diff
@@ -0,0 +1,187 @@
...
+  }
+}
+
+UploadSubstancesButton.propTypes = {
+  tooltip: PropTypes.string,
+  buttonTitle: PropTypes.string,
+  disabled: PropTypes.bool,
+};
+
+export default UploadSubstancesButton;
```
johandahlberg, line 187: I suppose since there is really no (major) state handling here there is no point of hooking it up to redux?

johandahlberg (approved):

src/clims/api/endpoints/substance_file_details.py
withrocks: Yes, this is from the original code. Seems that they had a particular auth per just this additional get parameter. I'll remove this as we'll revisit the whole auth story later.

src/clims/services/extensible.py
withrocks: I intended to write "may be" as "it's allowed to be" rather than "perhaps". Both forms exist. I think I'll just remove this as that should be hidden from the caller anyways.

src/sentry/static/sentry/app/views/substances/uploadSubstancesButton.jsx
withrocks: I think so, at least for the moment. It might fit better with the whole solution to have everything in redux at a later state, but it felt like an overkill here.

src/clims/api/endpoints/substance_file_details.py
withrocks: What I meant there is that when a file is uploaded, one can set the content type of it. Then when downloading it (like here), we should use the content-type that was used when it was uploaded.

Here however, for some reason the content-type of the file is not set (headers was also unexpectedly a list). I assume it's probably my fault after some refactoring. This is not a huge issue at the moment, since we can just download these files as `application/octet-stream`, but I assume there may be cases where we wouldn't want that (although I don't see it at the moment).

src/clims/api/endpoints/substance_file.py
withrocks: Do you mean pparam?

johandahlberg: Yes.

src/clims/api/endpoints/substance_file_details.py
johandahlberg: :+1:

src/clims/api/endpoints/substance_file_details.py
johandahlberg: :+1:

src/clims/api/endpoints/substance_file.py
withrocks: That is used in the codebase. They parse it here: src/sentry/utils/apidocs.py

This is a sentry thing. These are "path" parameters used in the api endpoint documentation. Look for "path parameter" here:

https://docs.sentry.io/api/events/get-project-group-index/

johandahlberg: Aha! Got it!

PR #61, opened 2019-10-17
253, pool domain object

GitEdvard:
Purpose:
To be able to instantiate pools with SubstanceBase. Pools are instantiated by calls similar to

mypool = Pool(name='mypool', organization=myorg, parents=[sample1, sample2])

Implementation:
If SubstanceBase is initialized with parents, these are stored internally in a cache until save() is called. SubstacneBase objects instantiated from a fetched db object (_wrapped_version), the parents property comes along with the input wrapped version.

Comments:
I removed the ancestry thing from ExtensibleBase. I thought that the _archetype.origins initialization got too fragmented, it would happen in both ExtensibleBase and in SubstanceBase.

### DIFF
--- a/src/clims/services/extensible.py
+++ b/src/clims/services/extensible.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 
 import six
-from clims.models import ExtensibleType, ExtensiblePropertyType, ExtensibleProperty
+from clims.models.extensible import ExtensibleType, ExtensiblePropertyType, ExtensibleProperty
 from django.db import transaction
 
 
@@ -107,12 +107,9 @@ class ExtensibleBase(object):
         self._property_bag = PropertyBag(self)
         self._name_before_change = None
         self._wrapped_version = kwargs.get("_wrapped_version", None)
-        self._has_ancestry = True
         if self._wrapped_version:
             self._archetype = self._wrapped_version.archetype
-
-        # Stop here if we're wrapping a version that already exists
-        if self._wrapped_version:
+            # Stop here if we're wrapping a version that already exists
             return
 
         name = kwargs.pop("name", None)
@@ -131,6 +128,10 @@ class ExtensibleBase(object):
         for key, value in six.iteritems(kwargs):
             setattr(self, key, value)
 
+    def _save_subclass_specifics(self, creating):
+        # override this!
+        pass
+
     @transaction.atomic
     def save(self):
         creating = self.id is None
@@ -138,11 +139,6 @@ class ExtensibleBase(object):
             self._archetype.save()
             self._wrapped_version.archetype = self._archetype
             self._wrapped_version.save()
-
-            if self._has_ancestry:
-                # We want the origin point(s) to always be populated, also for the origins themselves, in
-                # which case it points to itself. This way we can find all related samples in one query.
-                self._archetype.origins.add(self._archetype)
             self._property_bag.save(self._wrapped_version)
         else:
             # Updating
@@ -165,6 +161,8 @@ class ExtensibleBase(object):
             self._archetype.versions.add(new_version)
             self._property_bag.save(new_version)
             self._wrapped_version = new_version
+
+        self._save_subclass_specifics(creating)
 
     @property
     def id(self):
--- a/src/clims/services/substance.py
+++ b/src/clims/services/substance.py
@@ -324,7 +324,7 @@ class SubstanceBase(ExtensibleBase):
     WrappedVersion = SubstanceVersion
 
     def __init__(self, **kwargs):
-        self._cached_parents = self._fetch_parents(kwargs)
+        self._unsaved_parents = self._fetch_parents(kwargs)
         super(SubstanceBase, self).__init__(**kwargs)
 
     def _fetch_parents(self, kwargs):
@@ -377,15 +377,15 @@ class SubstanceBase(ExtensibleBase):
         return SubstanceAncestry(self)
 
     def _save_parents(self):
-        parents = [substance_base._wrapped_version for substance_base in self._cached_parents]
+        parents = [substance_base._wrapped_version for substance_base in self._unsaved_parents]
         self._archetype.parents.add(*parents)
-        self._archetype.depth = max([p.depth for p in self._cached_parents]) + 1
+        self._archetype.depth = max([p.depth for p in self._unsaved_parents]) + 1
         self._archetype.save()
 
     def _get_origins(self):
         origins = list()
-        if self._cached_parents:
-            for p in self._cached_parents:
+        if self._unsaved_parents:
+            for p in self._unsaved_parents:
                 for origin in p._archetype.origins.all():
                     origins.append(origin)
         else:
@@ -394,7 +394,7 @@ class SubstanceBase(ExtensibleBase):
 
     def _save_subclass_specifics(self, creating):
         if creating:
-            if self._cached_parents:
+            if self._unsaved_parents:
                 self._save_parents()
 
             # We want the origin point(s) to always be populated, also for the origins themselves, in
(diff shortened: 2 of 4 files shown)


src/clims/services/substance.py
```diff
@@ -113,19 +114,16 @@ def to_wrapper(self, model):
...
         try:
             SpecificExtensibleType = self._app.extensibles.get_implementation(
                 substance_version.archetype.extensible_type.name)
             return SpecificExtensibleType(_wrapped_version=substance_version, _app=self._app)
         except ExtensibleTypeNotRegistered:
-            # This is an unregistered instance. This can happen for example when we have
-            # an instance that used to be registered but the Python version has been removed
-            # or rename.
-            # We must use the base class to wrap it:
-            return SubstanceBase(_wrapped_version=substance_version, _unregistered=True)
```
johandahlberg: I think this should actually return `SubstanceBase` as before. The reason is that users need to be able to still retrieve information that was generated by plugins that are no longer installed.

tests/clims/models/test_substance.py
```diff
@@ -107,6 +107,64 @@ def do_asserts(substance):
...
+    def test_create_pool__from_two_aliquots__pool_origin_are_original_samples(self):
+        # Arrange
+        props = dict(preciousness='*o*', color='red')
+        sample1 = self.create_gemstone(**props)
+        sample2 = self.create_gemstone(**props)
+        aliquot1 = sample1.create_child()
+        aliquot2 = sample2.create_child()
+
+        # Act
+        pool = GemstoneSample(
```
johandahlberg: Perhaps we want to look at the terminology here? Aliquots and pools may not make that much sense for gemstones? Perhaps we want to call it something like `slice` and `combo` instead? While it seems like a minor thing, I think that keeping the terminology clear of liquid based lab stuff will help us avoid making assumptions that will make life difficult for us, later down the road.

src/clims/services/substance.py
```diff
@@ -291,8 +289,19 @@ class SubstanceBase(ExtensibleBase):
...
 
     def __init__(self, **kwargs):
+        self._cached_parents = self._fetch_parents(kwargs)
         super(SubstanceBase, self).__init__(**kwargs)
 
+    def _fetch_parents(self, kwargs):
+        wrapped_version = kwargs.get("_wrapped_version", None)
+        parents = kwargs.pop("parents", None)
+
+        if wrapped_version and parents:
```
johandahlberg, line 334: Perhaps the error message here should be more explanatory? Something along the lines of `A substance must either be created from a set of parents or a wrapped version`? Not sure that is exactly what the problem is, but I think that it is a good practices to let the AssertionError give a more though description of what went wrong.

johandahlberg (approved):
I had some minor comments/questions - but this looks good to me.

src/clims/services/substance.py
GitEdvard: Ok, I can put it back. But, I want to point out that this cost me some time in searching why a test was failed. I was forced to go into low level details, like the difference between substance-version and substance.

tests/clims/models/test_substance.py
GitEdvard: ok.

src/clims/services/substance.py
GitEdvard: ok!

src/clims/services/substance.py
johandahlberg: Yes. I see your point. :+1: Hopefully as we evolve the APIs this should happen less often. But while we are still in "under construction mode" I think that is something that we will have to deal with.

src/clims/services/extensible.py
```diff
@@ -163,6 +158,8 @@ def save(self):
             self._property_bag.save(new_version)
             self._wrapped_version = new_version
 
+        self._save_subclass_specifics(creating)
```
withrocks, line 165: Another approach here would be to have the subclass call if it wants the parents behaviour (which it does in our case). I think that might be easier to wrap ones head around.

src/clims/services/substance.py
```diff
@@ -291,8 +289,19 @@ class SubstanceBase(ExtensibleBase):
     WrappedVersion = SubstanceVersion
 
     def __init__(self, **kwargs):
+        self._cached_parents = self._fetch_parents(kwargs)
```
withrocks: Here one might argue (@johandahlberg :)) that this isn't actually a cache but rather a list of unsaved entries (as saying that it's a cache implies that it can be cleared without consequences).

src/clims/services/substance.py
withrocks: Maye we could turn around the order of operations a bit here. So if we first call super from `__init__`, super will have made sure to parse all parameters accordingly. Then the logic in _fetch_parents would be:

```
parents = kwargs.pop("parents", None)
if self._wrapped_version.id and parents:
   raise AssertionError('Substance can only be initialized with parents when creating a new instance')
```

Even better would probably be to have a property on the extensible similar to:

```
@property
def is_new(self):
    return self._wrapped_version.id is not None
```

But that can be refactored later!

src/clims/services/substance.py
```diff
@@ -330,6 +339,32 @@ def parents(self):
     def to_ancestry(self):
         return SubstanceAncestry(self)
 
+    def _save_parents(self):
+        parents = [substance_base._wrapped_version for substance_base in self._cached_parents]
+        self._archetype.parents.add(*parents)
+        self._archetype.depth = max([p.depth for p in self._cached_parents]) + 1
```
withrocks: Interesting! I knew of course that we have a graph but not a tree, but I still felt that depth would be a good name for this, probably because I first thought of this as being a tree rather than a graph. But since this is a graph, we should perhaps call this "max_distance" or "level". We can look into that later as the need arises.

tests/clims/models/test_substance.py
```diff
@@ -107,6 +107,63 @@ def do_asserts(substance):
...
+
+        assert 'xxx' == fetched.preciousness
+        assert 2 == len(fetched.parents)
+
+    def test_depth__with_multiple_parents_with_different_depths__new_depth_based_on_highest(self):
+        # I just picked a policy for this, I had to choose something
+        # Arrange
+        props = dict(preciousness='*o*', color='red')
+        substance1 = self.create_gemstone(**props)
+        substance1._archetype.depth = 4
```
withrocks, line 152: It would be good to expose the depth on the SubstanceBase, but I think for now we can wait with it, since it's unclear if it will be called depth.

withrocks (commented):
Looks great! I have a few comments, as always ;), but those are just for the future.

src/clims/services/extensible.py
GitEdvard: Yes, but I was uncertain about @transaction.atomic decorator in the base class. Maybe we could take a quick talk about it?

src/clims/services/substance.py
GitEdvard: Good point, I think _unsaved_parents might be better.

src/clims/services/substance.py
GitEdvard: Ok!

tests/clims/models/test_substance.py
GitEdvard: Maybe we could leave this test as it is? I think this test is informative to show how depth works, and what is the current status of depth concept. When depth is renamed, it would be easy to adjust this test?

src/clims/services/extensible.py
withrocks: Yes, it occurred to me that I thought exactly the same when I was thinking about this the other day! I think a solution to that is to add it also to the parent. But this is just food for thought I think.

tests/clims/models/test_substance.py
withrocks: Absolutely :)

PR #63, opened 2019-10-17
Group by Container/SampleType/Nothing in sample UI

johandahlberg:
Not the most beautiful js code the world ever saw (but then again I'm still trying to figure it out). This allows the sample list UI to aggregate by sample (no aggregation), container and sample type.

I did not put the view state of the container into redux now, since I don't think any other components will actually access it. I don't know what you think about that? Also perhaps I should write some tests for this? I'll get back to this later though, so if you like we can merge this before the demo tomorrow.

### DIFF
--- a/src/sentry/static/sentry/app/views/substances/substances.jsx
+++ b/src/sentry/static/sentry/app/views/substances/substances.jsx
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
 import {substancesGet} from 'app/redux/actions/substance';
-//import Switch from 'app/components/switch';
 import ReactTable from 'react-table';
 import 'react-table/react-table.css';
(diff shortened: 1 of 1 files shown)


src/sentry/static/sentry/app/views/substances/substances.jsx
```diff
@@ -70,19 +78,66 @@ class Substances extends React.Component {
...
+      },
+    ];
+
+    return (
+      <DropdownAutoComplete
+        items={groupedItems}
+        alignMenu="left"
+        onSelect={this.selectGroupBy.bind(this)}
+      >
+        {({isOpen, selectedItem}) => (
```
withrocks, line 113: It's interesting that they use this pattern, i.e. that there is a function for rendering each button.

withrocks (approved):
This looks good to me! I agree that we shouldn't need to add view state to the redux store. It should at least be a simple thing to fix later if it has some benefits.

PR #64, opened 2019-10-22
Clims 247 handle none value in sub base

johandahlberg:
This allows fields to be set to None (provided that they are nullable - which they are by default). I also moved the Field things into extensible, since it seemed more related to that than to substance.

### DIFF
--- a/tests/clims/models/test_substance.py
+++ b/tests/clims/models/test_substance.py
@@ -24,30 +24,59 @@ class SubstanceTestCase(TestCase):
 
 class SubstancePropertiesTestCase(TestCase):
 
+    class ExampleSample(SubstanceBase):
+        moxy = FloatField("moxy")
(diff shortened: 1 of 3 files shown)


tests/clims/models/test_substance.py
```diff
@@ -24,31 +24,60 @@ def register_gemstone_type(self):
...
+
+        sample = self.ExampleSample(name=name,
+                                    organization=self.organization,
+                                    moxy=None,
+                                    cool=cool,
+                                    erudite=erudite)
+        sample.save()
+
+        fetched_sample = self.app.substances.get(name=sample.name)
+        fetched_sample.moxy is None
```
withrocks: Missing asserts?

withrocks (approved):
I like it!

PR #66, opened 2019-10-23
Containers

withrocks:


src/clims/models/substance.py
```diff
@@ -30,8 +30,6 @@ def __init__(self, *args, **kwargs):
 
     project = FlexibleForeignKey('sentry.Project', null=True)
 
-    location = FlexibleForeignKey('clims.Location', null=True)
```
johandahlberg, line 33: Do I understand this correctly that the interpretation is that one substance can be in multiple containers at the same time? Does that make sense?

src/clims/services/container.py
```diff
@@ -1,13 +1,99 @@
...
+            return cls(*key)
+        else:
+            raise NotImplementedError("Can't use {} as an index".format(type(key)))
+
+
+class InvalidContainerIndex(Exception):
+    pass
+
+
+class PlateIndex(ContainerIndex):
```
johandahlberg, line 47: It this something that should go into a plugin later?

src/clims/services/container.py
```diff
@@ -1,13 +1,99 @@
...
+    @classmethod
+    def from_string(cls, key):
+        """
+        Given a human-readable index into the container, e.g. "A:1", creates a new index.
+        """
+        m = cls.STRING_PATTERN.match(key)
+        if not m:
+            raise InvalidContainerIndex(
+                "Can't interpret '{}' as a container location".format(key))
+        row = m.group('row').upper()
```
johandahlberg, line 80: Minor thing, but I don't know if we want to validate that the letter is in the A-Z range? Or is ":hear_no_evil: :1" a valid index?

src/clims/services/container.py
```diff
@@ -44,7 +127,143 @@ class ContainerBase(ExtensibleBase):
...
+    def to_string(self, compressed=False, short=False):
+        """
+        Returns a detailed representation of the container. This is more detailed than
+        __str__ provides.
+        """
+
+
+class PlateBase(ContainerBase):
+
+    TRAVERSE_BY_ROW = 1
```
johandahlberg, line 203: Should traverse orders be wrapped in an enum, maybe?

johandahlberg (approved):
I had a few comments - and I think it would be good if @GitEdvard also went over this since he will use the container stuff. I'll go ahead and merge this now though, so that I can keep working on the Project, and @GitEdvard can continue on the container things based on this.

PR #67, opened 2019-10-23
Clims 257 extendible projects

johandahlberg:
This adds the concept of projects to clims. There are still some things to do - e.g. writing tests for the project service, but since this also contains some general refactoring of how the other services work with regards to wrapping objects, so I thought it would be good idea to get this into the develop branch.

Part of this looks a bit odd, since I had to move some code around in the service class to get the interpreter to have things defined in the correct order. But all that really happened there is that I moved the wrapper methods to their own mixin.

This also fixes two minor bugs:
 - It removes the submodule that I accidentally added in previous PR
 - Fixes the create example data script to match up with the changes to the API.


src/clims/models/project.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    __core__ = True
+
+    def __init__(self, *args, **kwargs):
+        super(Project, self).__init__(*args, **kwargs)
+
+    name = models.TextField()
+    organization = FlexibleForeignKey('sentry.Organization', related_name='%(class)s_organization')
+
+    # TODO What other properties do we want to be native to the projects?
```
withrocks, line 25: I agree. I guess we can just drag things up from the properties as we progress, in particular after we've created the project subclass in commonlims-snpseq.

src/clims/models/substance.py
```diff
@@ -26,9 +26,9 @@ def __init__(self, *args, **kwargs):
 
     name = models.TextField()
 
-    organization = FlexibleForeignKey('sentry.Organization')
+    organization = FlexibleForeignKey('sentry.Organization', related_name='%(class)s_organization')
```
withrocks: Wouldn't the related_name just be `substances`? That would mean that if you have an Organization, you can query for the substances using this syntax:

```
org = Organization.get(name=lab)
org.substances.all()
```

As this is now (if I'm not missing something) it would be:

```
org.substances_organization.all()
```

src/clims/services/container.py
```diff
@@ -267,3 +238,16 @@ def to_table():
         if header:
             rows.insert(0, self.name)
         return "\n".join(map(six.text_type, rows))
+
+
+class ContainerService(WrapperMixin):
```
withrocks: This is great. I believe we should also inherit from object, since we're still using Python 2:

https://stackoverflow.com/a/38140423/282024

src/clims/services/project.py
```diff
@@ -0,0 +1,49 @@
...
+    def __init__(self, **kwargs):
+        super(ProjectBase, self).__init__(**kwargs)
+
+    def _to_wrapper(self, model):
+        """
+        Wraps either a SubstanceVersion or Substance as a higher-level object
+        """
+        return self._app.projects.to_wrapper(model)
+
+    @property
```
withrocks, line 23: I guess we would later also want to add these to a mixin.

withrocks (commented):
Looks great! I wonder if you want to change that related_name before merging.

src/clims/models/project.py
johandahlberg: :+1:

src/clims/models/substance.py
johandahlberg: I think you are right. I'll fix this.

src/clims/services/container.py
johandahlberg: Good catch!

src/clims/services/project.py
johandahlberg: Your thinking of the properties? Yes, I suppose so. Or be moved into the `ExtensibleBase`?

tests/clims/models/test_project.py
```diff
@@ -0,0 +1,57 @@
...
+class TestProject(TestCase):
+
+    def setUp(self):
+        self.register_extensible(VampireFangStudyProject)
+        self.register_extensible(VampireFangSample)
+
+    def test_can_create_project(self):
+        project = VampireFangStudyProject(name="project1", organization=self.organization)
+        project.save()
+        Project.objects.get(name=project.name)  # Raises DoesNotExist if it wasn't created
```
GitEdvard: I'm concerned about using Project class directly in tests. I think the tests should show a reader how to use the classes, and the class Project should not be used by upper layers, it should be wrapped (and hidden) within ProjectBase.

src/clims/services/wrapper.py
```diff
@@ -0,0 +1,35 @@
...
+class WrapperMixin(object):
+    """
+    This mixin can be used to provide wrapping methods to a service class which
+    handles translating models of `ExtensibleModel` to their specific `ExtensibleBase`
+    implementations.
+    The implementing class needs to provide `_archetype_version`, `_archetype`, and
+    `_archetype_base`.
+    """
+
+    _archetype_version = None
```
GitEdvard: Perhaps it could be indicated that these are class variables and not instances of some class?

GitEdvard:
Now I have also reviewed this PR. Nice work! I like the wrapper mixin base class.

GitEdvard (approved):

src/clims/services/extensible_service_api.py
```diff
@@ -0,0 +1,45 @@
...
+        pass
+
+    def get(self, **kwargs):
+        get_args = {}
+        for key, value in kwargs.items():
+            get_args['archetype__{}'.format(key)] = value
+        if 'latest' not in get_args.keys():
+            get_args['latest'] = True
+
+        substance_model = self._archetype_version_class.objects.prefetch_related('properties').get(**get_args)
```
GitEdvard: I think it should not be substance_model :)

src/clims/services/extensible_service_api.py
```diff
@@ -0,0 +1,45 @@
...
+        return self._archetype_version_class.objects.filter(latest=True).prefetch_related('properties')
+
+    def filter(self, *args, **kwargs):
+        pass
+
+    def get(self, **kwargs):
+        get_args = {}
+        for key, value in kwargs.items():
+            get_args['archetype__{}'.format(key)] = value
+        if 'latest' not in get_args.keys():
```
GitEdvard, line 41: If I understand 'not in \<list\>' correctly, it should only get true for exact matches. All keys in get_args begins with 'archetype'

src/clims/services/extensible_service_api.py
```diff
@@ -0,0 +1,45 @@
...
+
+        return self._archetype_version_class.objects.filter(latest=True).prefetch_related('properties')
+
+    def filter(self, *args, **kwargs):
+        pass
+
+    def get(self, **kwargs):
+        get_args = {}
+        for key, value in kwargs.items():
+            get_args['archetype__{}'.format(key)] = value
```
GitEdvard: The word 'archetype__' is hard coded here. Shouldn't it be replaced by the different sub type names?

tests/clims/models/test_project.py
```diff
@@ -16,7 +15,7 @@ def setUp(self):
     def test_can_create_project(self):
         project = VampireFangStudyProject(name="project1", organization=self.organization)
         project.save()
-        Project.objects.get(name=project.name)  # Raises DoesNotExist if it wasn't created
+        self.app.projects.get(name=project.name)  # Raises DoesNotExist if it wasn't created
```
GitEdvard, line 18: I would like it to return None if it doesn't exist!

GitEdvard (commented):
Super nice to remove a lot of duplication! I got a few comments. Tell me when you want to merge.

src/clims/services/extensible_service_api.py
johandahlberg: Damn you copy-paste! Why don't you automatically re-name stuff? :wink:

tests/clims/models/test_project.py
johandahlberg: I suggest we keep the API as it is for now. But later, that might be what we want.

src/clims/services/extensible_service_api.py
johandahlberg: No, because they all have a related field called archetype, which is what we tap into here.

src/clims/services/extensible_service_api.py
johandahlberg: Good point. Fix this now.

src/clims/services/extensible_service_api.py
GitEdvard: Ok, good.

tests/clims/models/test_project.py
GitEdvard: ok

PR #69, opened 2019-10-24
260, Add option to read excel as read only

GitEdvard:
Purpose:
Opening excel files in read only mode is a lot faster than ordinary mode. 

Implementation:
The NamedTemporyFile is moved from the as_excel() method to the outer context. Apparently, the temp file is needed when reading from the excel file later on. In normal mode, the temp file is not deleted outside the "with ..." context (I thought that it was deleted and worked anyway). In read only mode the file is deleted, and the script stopped working. 

Now, the temp file is created within the Handler class (see test), and removed in the destructor method.

### DIFF
--- a/tests/clims/services/test_substances.py
+++ b/tests/clims/services/test_substances.py
@@ -78,6 +78,14 @@ class TestGemstoneSampleSubmission(SubstanceTestCase):
 
 
 class MyHandler(SubstancesSubmissionHandler):
+    def __init__(self, context, app):
+        super(MyHandler, self).__init__(context, app)
+        from tempfile import NamedTemporaryFile
+        self.temp_file = NamedTemporaryFile(suffix='.xlsx', delete=False)
+
+    def __del__(self):
+        os.remove(self.temp_file.name)
+
     def handle(self, file_obj):
         csv = self._as_csv(file_obj)
         for line in csv:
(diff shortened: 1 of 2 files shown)


tests/clims/services/test_substances.py
```diff
@@ -78,6 +78,14 @@ def test_gemstone_submission_handler__with_xlsx__3_samples_found(self):
 
 
 class MyHandler(SubstancesSubmissionHandler):
+    def __init__(self, context, app):
+        super(MyHandler, self).__init__(context, app)
+        from tempfile import NamedTemporaryFile
+        self.temp_file = NamedTemporaryFile(suffix='.xlsx', delete=False)
+
+    def __del__(self):
+        os.remove(self.temp_file.name)
```
johandahlberg: I already merged the PR in the snpseq plugin related to this, which was perhaps not so smart of me, since now that I look at this I'm not 100% if this is something that could come back an cause us trouble in the future. Basically my questions boils down to if we can trust the Handlers to have a life cycle in which the constructor will be called at the appropriate time.

Consider the scenario where we instantiate a handler class is created and then keep alive for the duration of the application (like we would typically do for e.g. services). It this case the destructor would never be called. Do you think that this is going to be a problem in this case or not? I'm unsure, but I wanted to raise the question.

johandahlberg (approved):

PR #70, opened 2019-10-28
clims 263, Suggestion for handling temp files

GitEdvard:
Purpose:
This is a sketch for handling temp files when reading from excel. 

Implementation:
OrganizationFile is here wrapped within ExcelFileWrapper in order to handle the temp file. ExcelParser (in snpseq) is then not to be a context manager any longer. The good thing about this implementation is that temp file is encapsulated within a single class, and the intended use for that class is as a context manager. 

Comments:
I find it awkward to make OrganizationFile a context manager (in order to get rid of the wrapper class). How about this call
        org_file = OrganizationFile.objects.create(
when org_file is a context manager?

I think it shouldn't be named ExcelBlabla. load_file() is so generic, and it looks strange with that name here. My suggestion is to call it something generic, and include implementation of as_csv.

Another thought:
we could instantiate ExcelFileWrapper within handler, when it's time to extract excel. Problem here though, I think it might be hard to find this wrapper for a plugin developer.

### DIFF
--- a/src/clims/migrations/0001_initial.py
+++ b/src/clims/migrations/0001_initial.py
@@ -172,7 +172,6 @@ class Migration(migrations.Migration):
             options={
                 'db_table': 'clims_organizationfile',
             },
-            bases=(clims.models.file.StructuredFileMixin, models.Model),
         ),
         migrations.CreateModel(
             name='PluginRegistration',
--- a/src/clims/models/file.py
+++ b/src/clims/models/file.py
@@ -1,9 +1,45 @@
 from __future__ import absolute_import
-
+import os
 from django.db import models
-
+from tempfile import NamedTemporaryFile
 from sentry.db.models import FlexibleForeignKey, Model, sane_repr
 from openpyxl import load_workbook
+
+
+class ExcelFileWrapper(object):
+    """
+    Wraps an organization file in order to produce an excel file.
+    This functionality is wrapped in its own class because excel handling
+    needs a temporary file that needs to be removed after usage.
+    """
+
+    def __init__(self, organization_file):
+        self.temp_file = None
+        self.organization_file = organization_file
+        self.name = organization_file.name
+
+    def __enter__(self):
+        self.temp_file = NamedTemporaryFile(suffix='.xlsx')
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        os.remove(self.temp_file.name)
+
+    def as_excel(self, read_only=True):
+        """
+        Returns the file as an excel file. The caller has the responsibility to
+        delete the temp file after usage!
+        """
+        with open(self.temp_file.name, 'wb') as f:
+            for _, chunk in enumerate(self.organization_file.file.getfile()):
+                f.write(chunk)
+
+        workbook = load_workbook(self.temp_file.name, data_only=True, read_only=read_only)
+
+        return workbook
+
+    def as_csv(self):
+        return self.organization_file.as_csv()
 
 
 class StructuredFileMixin(object):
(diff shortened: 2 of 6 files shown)


tests/clims/services/test_substances.py
```diff
@@ -32,10 +34,13 @@ def csv_sample_submission_path():
 class TestGemstoneSampleSubmission(SubstanceTestCase):
     def setUp(self):
         self.gemstone_sample_type = self.register_gemstone_type()
+        logger = logging.getLogger('clims.files')
```
johandahlberg, line 37: Don't know if logging should be enabled in the tests?

johandahlberg (commented):
I think we are moving in the right direction now. I do think however that to keep with intention of the original `StructuredFileMixin` we should get one class to handle all types of files csv/xlsx/xml/whatever.

So that we could do something like this:
```
        with StructuredFileWrapper(org_file) as wrapped_org_file:
            plugins.handle(SubstancesSubmissionHandler, context, True, wrapped_org_file)
```

And in the handler you could then call `as_excel`, `as_csv`, etc on the file as needed. And temporary files could be used as needed? Does this make sense to you?

tests/clims/services/test_substances.py
GitEdvard: This is my trick to get rid of the logging!

GitEdvard:
Sounds great. I will make some changes so please dont merge yet.

GitEdvard:
Now I have an updated suggestion. ExcelFileWrapper has changed name to MultiFormatFile, and now includes all methods from StructuredFileMixin. OrganizationFile only inherits from model.

src/clims/models/file.py
```diff
@@ -1,27 +1,42 @@
...
 
-    def as_excel(self, temp_file, read_only=True):
-        """
-        Returns the file as an excel file. The caller has the responsibility to
-        delete the temp file after usage!
-        """
-        with open(temp_file.name, 'wb') as f:
-            for _, chunk in enumerate(self.file.getfile()):
+    def __init__(self, organization_file):
+        self.temp_file = None
```
johandahlberg: A though there - perhaps we should make all the fields except `name` with `_`, since I'm guessing that they are not supposed to be accessed by the library user?

src/clims/migrations/0001_initial.py
```diff
@@ -172,7 +172,6 @@ class Migration(migrations.Migration):
             options={
                 'db_table': 'clims_organizationfile',
             },
-            bases=(clims.models.file.StructuredFileMixin, models.Model),
```
johandahlberg, line 175: Just a note that in the future we shouldn't change old migrations, since it means we can run them over existing data. At this stage we can just reset the db, though, so it shouldn't be a problem.

johandahlberg (approved):
LGTM. One small comment on making some fields "private", but other than that I have nothing to change. I think this was a good discussion, and I think it has made the APIs clearer.

src/clims/migrations/0001_initial.py
GitEdvard: make reset-db actually failed when the class StructuredFileMixin was removed. If it were existing data in db, it seems that we were to keep an empty class named StrucutedFileMixin in order to the migration to succed?

src/clims/models/file.py
GitEdvard: Yes, I agree

src/clims/migrations/0001_initial.py
johandahlberg: I'm not sure I understand. Did you create a new migration script?

PR #71, opened 2019-10-29
clims 265, fix navigation property between project and substance

GitEdvard:
Purpose:
In order to initiate samples with a project when importing sample submission.

Implementation:
Change the project property on substance so that it take advantage of the foreign key for django-model base class for substance. 

https://github.com/commonlims/commonlims/blob/7b6232855db802626810e3a3fe640184e2f464f5/src/clims/models/substance.py#L31

Before, project was located in the property-bag, and thus declared as an extensible property. 

Comments:
I decorated project property with lazyprop, in order to avoid excessive round trips to db in plugins. 

Comment #2:
This call will not work:
project = GetProjectSomehow("myproject")
self.app.substances.get(project=project)
nor this
self.app.substances.get(project="myproject")
I think I write a ticket for it.

### DIFF
--- a/src/clims/services/substance.py
+++ b/src/clims/services/substance.py
@@ -131,8 +131,7 @@ class SubstanceBase(HasLocationMixin, WrapperMixin, ExtensibleBase):
 
     def __init__(self, **kwargs):
         self._check_parameters(kwargs)
-        self._unsaved_parents = kwargs.pop("parents", None)
-        self._unsaved_project = kwargs.pop('project', None)
+        self._unsaved_parents = kwargs.pop('parents', None)
         super(SubstanceBase, self).__init__(**kwargs)
         self._new_location = None
 
@@ -199,6 +198,7 @@ class SubstanceBase(HasLocationMixin, WrapperMixin, ExtensibleBase):
         parents = [substance_base._wrapped_version for substance_base in self._unsaved_parents]
         self._archetype.parents.add(*parents)
         self._archetype.depth = max([p.depth for p in self._unsaved_parents]) + 1
+        self._archetype.save()
 
     def _get_origins(self):
         origins = list()
@@ -214,11 +214,6 @@ class SubstanceBase(HasLocationMixin, WrapperMixin, ExtensibleBase):
         if creating:
             if self._unsaved_parents:
                 self._save_parents()
-
-            if self._unsaved_project:
-                self._archetype.project = self._unsaved_project._archetype
-
-            self._archetype.save()
 
             # We want the origin point(s) to always be populated, also for the origins themselves, in
             # which case it points to itself. This way we can find all related samples in one query.
--- a/src/clims/services/substance.py
+++ b/src/clims/services/substance.py
@@ -17,6 +17,7 @@ from clims.models.file import OrganizationFile
 from clims.handlers import SubstancesSubmissionHandler, HandlerContext
 from clims.services.wrapper import WrapperMixin
 from clims.services.extensible_service_api import ExtensibleServiceAPIMixin
+from clims.utils import lazyprop
 
 
 class NotFound(Exception):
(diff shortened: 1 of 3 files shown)


johandahlberg:
Looking at this I think that perhaps it can be made simpler. I added your tests to my own branch and poked around a bit, and except for the laziness part I think that the changes indicated below should be be sufficient to solve this:

```
--- a/src/clims/services/substance.py
+++ b/src/clims/services/substance.py
@@ -152,6 +152,14 @@ class SubstanceBase(HasLocationMixin, WrapperMixin, ExtensibleBase):
         """
         return self._app.substances.to_wrapper(model)

+    @property
+    def project(self):
+        return self._app.projects.to_wrapper(self._archetype.project)
+
+    @project.setter
+    def project(self, project):
+        self._archetype.project = project._archetype
``` 

I might be missing something here, but provided I am not I think I this solution since it brings all the code for getting and setting the property together in a more obvious way. This will trigger database round trips when we get the project, but I think that we can solve that by having the framework pre-fetch the project when it gets the substance. Since the amounts of data read should still be relatively minor I think that is a more elegant solution.

Does this look reasonable to your? Or am I missing something here?

GitEdvard:
Yes, it is more straightforward now. I think it would be nice to let the plugin programmer have control of the life time of these navigation properties. Then he could avoid either unnecessary round trips or not up-to-date data.

src/clims/services/substance.py
```diff
@@ -130,21 +130,25 @@ class SubstanceBase(HasLocationMixin, WrapperMixin, ExtensibleBase):
...
+        has_specific_parameteres = name or organization or project or parents
+        if has_specific_parameteres and wrapped_version:
             raise AssertionError(
                 'A substance may either be instantiated from a wrapped version '
                 '(i.e. an already created object fetched from db), or '
-                'a new set of of parents. This call had both!')
-
-        return parents
+                'a new set of specific parameters, like name, project and so forth. '
+                'This call has both!')
```
johandahlberg: I think that with the property approach, there is no need to check this.

src/clims/services/substance.py
```diff
@@ -204,6 +215,11 @@ def _save_custom(self, creating):
             if self._unsaved_parents:
                 self._save_parents()
 
+            if self._unsaved_project:
+                self._archetype.project = self._unsaved_project._archetype
+
+            self._archetype.save()
+
```
johandahlberg: Same as the above goes for this. The property should take care of that.

johandahlberg (commented):
I'd suggest removing more code, as per my selections.

johandahlberg:
Regarding your comment on not being able to use the get method, what you need to do is something like: `self.app.substances.get(project__name=project.name)`. This is not super intuitive, and perhaps we want to add more utility methods, like `get_by_project_name`, to make this more accessible to API users. However, for now it's good to know that is the way to query across joined tables.

GitEdvard:
Thanks for the comments! Here is my thoughts on the check-method that was mentioned above:
Even without project as an input to init(), there is still two legitimate sets of parameters, either _wrapped_version, or (name, organization, parents, + properties). I would prefer two factory methods for this, but it's kind of complicated, because of the to_wrapper() methods. I wanted to express this in code, although plain text (in the assertion) is not the first choice.

GitEdvard:
Now I got a second thought about the initalization of project. For a plugin developer, it might be intuitive to give the call
sample = SqarkSample(name, organization, project)
the project would then be initalized as an extensible property and it will just not work for him. He must know in advance that 
sample = SqarkSample(name, organization)
sample.project = myproject
is the proper way to initialize sample.

johandahlberg:
I'm not sure I follow. If I understand your I wrote this test:

```
    def test_can_get_project_for_substance(self):
        project = VampireFangStudyProject(name="project1", organization=self.organization)
        project.save()
        sample = VampireFangSample(name="sample1", organization=self.organization, project=project)
        sample.save()
        sample_by_project = self.app.substances.get(project__name=project.name)
        assert sample_by_project.name == sample.name

    def test_can_get_project_for_substance_2(self):
        project = VampireFangStudyProject(name="project1", organization=self.organization)
        project.save()
        sample = VampireFangSample(name="sample1", organization=self.organization)
        sample.project = project
        sample.save()
        sample_by_project = self.app.substances.get(project__name=project.name)
        assert sample_by_project.name == sample.name
```
And it seems to work, which indicates to me that it should be alright either way. So I'm not sure that the `SqarkSample(name, organization, project)` syntax would instantiate the project as a property. The test would indicate that it is actually added to the project field (i.e. connecting it via the foreign key to the project table). The interaction between `setattr` and `property` is a little bit unclear to me, though, so I might be missing something here.

GitEdvard:
Ok, I misunderstood. I thought it wouldn't work with project as arg to init() method. I have a new commit.

GitEdvard:
Ok, now I'm ready. I think I leave the handling of this:
self.app.substances.get(project__name=project.name)
into a new PR. This PR is so long now. substances.get is supposed to return a single instance, which is probably not what we want here.

PR #72, opened 2019-10-31
clims 266, add method to filter substances on basis of project

GitEdvard:
Purpose:
The method to filter substances on basis of project is added because of a supposed future need. In addition, the framework is extended by implementing the filter() method on extensibles. Properties for extensibles are fetched in a way to prevent redundant db calls. Introduce a new type of test that test number of db calls for an operation. 

Comments:
* I had problems to set the django DEBUG setting at setup time. That's why I set it in the test. 
* Perhaps the test for number of db calls should be set in a separate test suite? Right now there is only one though.

### DIFF
--- a/src/clims/services/extensible.py
+++ b/src/clims/services/extensible.py
@@ -364,9 +364,13 @@ class PropertyBag(object):
             new_value = self.new_values.get(key, None)
             if new_value:
                 return new_value
-            persisted_value = self.extensible_wrapper._wrapped_version.properties.get(
-                extensible_property_type__name=key)
-            return persisted_value.value
+
+            # properties.all() is used to leverage the .prefetch_related() call
+            # that is used when fetching the extensible
+            properties = self.extensible_wrapper._wrapped_version.properties.all()
+            value = next(prop.value for prop in properties if prop.extensible_property_type.name == key)
+
+            return value
         except ExtensibleProperty.DoesNotExist:
             return None
 
--- a/src/clims/services/extensible.py
+++ b/src/clims/services/extensible.py
@@ -357,7 +357,8 @@ class PropertyBag(object):
             except ExtensibleProperty.DoesNotExist:
                 create(key, value)
 
-        self.new_values.clear()
+        # Note, new_values retained after save. Alternative would be to fetch from db and
+        # re-initiate.
 
     def __getitem__(self, key):
         try:
@@ -365,9 +366,9 @@ class PropertyBag(object):
             if new_value:
                 return new_value
 
-            # properties.all() is used to leverage the .prefetch_related() call
+            # '.all_properties' is used to leverage the .prefetch_related() call
             # that is used when fetching the extensible
-            properties = self.extensible_wrapper._wrapped_version.properties.all()
+            properties = self.extensible_wrapper._wrapped_version.all_properties
             value = next(prop.value for prop in properties if prop.extensible_property_type.name == key)
 
             return value
--- a/src/clims/services/extensible_service_api.py
+++ b/src/clims/services/extensible_service_api.py
@@ -31,18 +31,25 @@ class ExtensibleServiceAPIMixin(object):
 
         return self._archetype_version_class.objects.filter(latest=True).prefetch_related('properties')
 
-    def filter(self, *args, **kwargs):
-        pass
+    def filter(self, **kwargs):
+        get_args = self._get_filter_arguments(**kwargs)
+        models = self._archetype_version_class.objects.prefetch_related('properties').filter(**get_args)
+        return [self.to_wrapper(m) for m in models]
 
     def get(self, **kwargs):
-        # TODO I'm not at all sure that this is a perfect method to handle this.
-        #      but for now it might be suff
+        get_args = self._get_filter_arguments(**kwargs)
+        model = self._archetype_version_class.objects.prefetch_related('properties').get(**get_args)
+        return self.to_wrapper(model)
+
+    def _get_filter_arguments(self, **kwargs):
         get_args = {}
         if 'latest' not in get_args.keys():
             get_args['latest'] = True
         for key, value in kwargs.items():
-            if not key == 'latest':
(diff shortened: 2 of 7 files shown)


src/clims/services/extensible.py
```diff
@@ -364,9 +364,13 @@ def __getitem__(self, key):
             new_value = self.new_values.get(key, None)
             if new_value:
                 return new_value
-            persisted_value = self.extensible_wrapper._wrapped_version.properties.get(
-                extensible_property_type__name=key)
-            return persisted_value.value
+
+            # properties.all() is used to leverage the .prefetch_related() call
+            # that is used when fetching the extensible
+            properties = self.extensible_wrapper._wrapped_version.properties.all()
```
johandahlberg: This is neat. I actually took the time to read up on `prefetch_related` now, and there seems to be a lot of things to keep in mind there in order to get optimal performance. However, that is something that we can profile as we go, I think. With that said I think that this is perhaps a good candidate for using the `to_attr` functionality? Then we can use the properties directly without having to search them, and we can utilize the fact that is is a dict to get a O(1) look up here instead of searching through the properties for the correct key, which I suppose is O(N). You can find info about using `to_attr` here: https://docs.djangoproject.com/en/2.2/ref/models/querysets/#prefetch-related (you'll need to scroll down a bit though to find it).

src/clims/services/extensible_service_api.py
```diff
@@ -31,18 +31,29 @@ def _all_qs(self):
...
         get_args = {}
         if 'latest' not in get_args.keys():
             get_args['latest'] = True
         for key, value in kwargs.items():
-            if not key == 'latest':
+            if key == 'project':
+                get_args['archetype__project__name'] = value.name
+            elif key == 'project_name':
+                get_args['archetype__project__name'] = value
+            elif not key == 'latest':
```
johandahlberg, line 60: I believe that we are missing a final `else` clause here.

tests/clims/models/test_substance.py
```diff
@@ -318,3 +317,31 @@ def test_can_iterate_through_all_versions_given_an_object(self):
 
         versions = [(s.version, s.properties) for s in sample.iter_versions()]
         assert len(versions) == 2
+
+    def test_fetch_property_after_instansiation__no_db_calls(self):
```
johandahlberg, line 321: Very good idea to add a test for this! 👍

johandahlberg (approved):
Looks good. Let me know what you think about the comments, and then we can get on to merging this.

src/clims/services/extensible.py
GitEdvard: Cool, I'll look through it

src/clims/services/extensible_service_api.py
GitEdvard: If you look to the original code, the original statement was, if not key == latest. It this was not fullfilled, nothing happened. Handling of 'latest' is a bit weird though, but I decided to wait with that.

tests/clims/models/test_substance.py
GitEdvard: Thanks.

src/clims/services/extensible_service_api.py
```diff
@@ -31,18 +31,29 @@ def _all_qs(self):
...
-        #      but for now it might be suff
+        get_args = self._get_filter_arguments(**kwargs)
+        model = self._archetype_version_class.objects.\
+            prefetch_related('properties__extensible_property_type').\
+            get(**get_args)
+        return self.to_wrapper(model)
+
+    def _get_filter_arguments(self, **kwargs):
         get_args = {}
         if 'latest' not in get_args.keys():
```
johandahlberg, line 53: I just noticed, I suppose it should be `kwargs` on this line?

johandahlberg: That is, this has been a bug all along...

src/clims/services/extensible_service_api.py
johandahlberg: Yeah. I think that this code is something that we will have to do some additional thinking on, to figure out what is the best approach here. For now I think this does the job. Perhaps it would be easier to understand if we added an `else` clause and added a comment that this is the `latest` case, and that we should just pass that? Or perhaps it is easier to do something like:

```
if key == 'latest:
    pass
elif key == 'project':
 [...]
else:
     get_args['archetype__{}'.format(key)] = value
```

johandahlberg (commented):
I'm not sure I understand why the changes in `substance.py` are necessary. I think that if possible we should move away to from these special handling of different variables. And since this worked before I'm wondering what prompted these changes?

Good work on the new pre-fetching. It is easier to understand how to access the properties without triggering a db call now.

GitEdvard:
Hello, I actually pushed a bit premature, before I was ready. And I have some explanations to give on why I made certain changes....

It was obviously so that the setter-thing for project didn't work. Don't ask me why I didn't discovered it earlier. It might be a missing make clean, or, it suddenly seems to go trolls in my monkey patching in sample submission tests, or it was me getting a touch of senility. 

When running the tests now, with the project setter implementation, there is an error message complaining about that self is NoneType. I have experienced similar errors in C#, when calling methods from subclasses within the base classes constructor. I can't tell if that comparison is relevant. But there is an error message. That's why I did go back to the previous implementation.

I also have a comment on the to_attr thing. I couldn't figure out how to use to_attr in order to avoid looping python objects, as you mention above. Implementing to_attr as I have in this commit is not super useful. I included it just to have it written down how to use it for later use. It has the side effect though, that all extensibles must be fetched from extensible service get() or filter() methods. For example Sustances.objects.get(), then to wrapper(), will lead to exceptions, since the attribute "all_properties" is not recognized. 

I was planning to fix handling of "latest" in extensible service, but there are trolls in my computer just now. I have to struggle for everything it seems. I wanted to set a bench mark with this PR before I go to vacation. I was planning to leave before lunch today. Argh.

PR #75, opened 2019-11-07
Plugin manager searches for implementations recursively

johandahlberg:
This should make the plugin manager search recursively for handlers - meaning that they don't all have to go in the `handler.py` file in the plugin anymore. Hopefully it also made the plugin manager code a little bit easier to follow, though I still think that we need more iterations on it to make sure it is easier to work with.

### DIFF
--- a/src/sentry/plugins/base/manager.py
+++ b/src/sentry/plugins/base/manager.py
@@ -263,9 +263,8 @@ class PluginManager(InstanceManager):
                 return clz
             elif nbr_of_subclasses > 1:
                 raise MultipleHandlersNotAllowed(
-                    "Trying to register the handler '{}' as the unique implementation "
-                    "of '{}' but already have '{}' registered".format(
-                        impl, baseclass, impl_already_reg))
+                    "Found the implementations: {}, when trying " +
+                    "to register a unique handler.".format(",".join(sub_classes)))
             else:
                 # We only had a single subclass, recurse to see if that hs any subclasses
                 return find_unique_deepest_impl(sub_classes[0])
--- a/src/sentry/plugins/base/manager.py
+++ b/src/sentry/plugins/base/manager.py
@@ -29,11 +29,6 @@ logger = logging.getLogger('clims.plugins')
 class PluginManager(InstanceManager):
     def __init__(self, class_list=None, instances=True):
         super(PluginManager, self).__init__(class_list, instances)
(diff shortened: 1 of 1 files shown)


src/sentry/plugins/base/manager.py
```diff
@@ -39,11 +44,9 @@ def __len__(self):
         return sum(1 for i in self.all())
 
     def register_handler_baseclasses(self):
-        from clims import handlers
-
-        for _name, cls in inspect.getmembers(handlers, inspect.isclass):
-            if cls != handlers.Handler and issubclass(cls, handlers.Handler):
-                self.handlers[cls] = set()
+        handler_subclasses = Handler.__subclasses__()
```
withrocks, line 42: Cool!

src/sentry/plugins/base/manager.py
```diff
@@ -26,6 +29,8 @@
 class PluginManager(InstanceManager):
     def __init__(self, class_list=None, instances=True):
         super(PluginManager, self).__init__(class_list, instances)
+        # TODO I guess work batches should be removed from here?
```
withrocks: Yes!

src/sentry/plugins/base/manager.py
```diff
@@ -248,41 +251,36 @@ def load_handlers(self, cls):
...
+        def find_unique_deepest_impl(clz):
+            sub_classes = clz.__subclasses__()
+            nbr_of_subclasses = len(sub_classes)
+            if nbr_of_subclasses == 0:
+                return clz
+            elif nbr_of_subclasses > 1:
+                raise MultipleHandlersNotAllowed(
+                    "Trying to register the handler '{}' as the unique implementation "
+                    "of '{}' but already have '{}' registered".format(
+                        impl, baseclass, impl_already_reg))
```
withrocks: `impl_already_reg` is not available here. Guess it should now be something like "Trying to ... but found <sub_classes>". Like the solution!

withrocks (changes requested):
Looks great! There is just that thing about `impl_already_reg`

src/sentry/plugins/base/manager.py
johandahlberg: That what happens when you copy-paste (and don't write tests).

withrocks (approved):

PR #76, opened 2019-11-08
Sample list enhanchments

withrocks:
Allows for simple search in the sample list view. The groupby is temporarily disabled, it waits on the implementation of the container endpoint to be fully implemented. Move towards a generic list component.

Includes some additional changes:
- Update React and react-table
- Adding UNSAFE_ as a prefix to componentWillMount. We will refactor this later. If not renamed like this, react throws a wall of warnings at us during build time.
- Some minor things


src/clims/api/endpoints/container.py
```diff
@@ -0,0 +1,36 @@
...
+from sentry.api.bases.organization import OrganizationEndpoint
+from clims.api.serializers.models.container import ContainerSerializer
+
+
+class ContainerEndpoint(OrganizationEndpoint):
+    authentication_classes = DEFAULT_AUTHENTICATION
+    permission_classes = (IsAuthenticated, )
+
+    def get(self, request, organization):
+        # TODO-nomerge: Filter by the organization
```
johandahlberg: TODO with nomerge here, so I guess we should actually fix this now?

src/clims/services/container.py
```diff
@@ -251,3 +251,22 @@ def __init__(self, app):
 
     def get_by_name(self, name):
         return self.get(name=name)
+
+    def _search_qs(self, query):
```
johandahlberg, line 255: Perhaps it would be worth adding some TODO comment here about the search here being a temporary thing?

tests/clims/models/test_container.py
```diff
@@ -9,9 +9,19 @@
...
 
 
 class TestContainer(TestCase):
+    def create_container_with_samples(self, prefix="container", sample_count=10):
+        container = self.create_container(GemstoneContainer, prefix=prefix)
+        for _ in range(sample_count):
+            sample = self.create_substance(GemstoneSample, color='red')
+            container.append(sample)
+        container.save()
+        return container
```
johandahlberg: This duplicates the code in `ContainerTest`

tests/clims/models/test_container.py
```diff
@@ -133,6 +143,21 @@ def test_can_output_default_string_info(self):
...
+    def test_get_content_incurs_no_extra_call(self):
+        # Getting the contents from a container incurs no extra calls to the DB if called correctly
+
+        from django.db import connection
+        q1 = len(connection.queries)
+        for ix in range(1):
+            container = self.create_container_with_samples(
+                prefix="container-{}".format(ix), sample_count=1)
+            container.save()
+        q2 = len(connection.queries)
```
johandahlberg, line 149: No assertion here.

src/clims/api/endpoints/substance.py
```diff
@@ -27,24 +30,19 @@ def handle_results(qs):
...
-                    json['properties']['volume'] = random.randint(0, 500)
-                if 'sample_type' not in json['properties']:
-                    json['properties']['sample_type'] = random.choice(["Amplicon", "WGS", "WES"])
-                if 'priority' not in json:
-                    json['priority'] = random.randint(1, 10)
-                if 'days_waiting' not in json:
-                    json['days_waiting'] = random.randint(0, 100)   # NOTE: We would have created and origin_created instead
-                if 'container' not in json['properties']:
-                    json['properties']['container'] = random.choice(["container 1", "container 2", "container 3"])
+                json['position'] = {'index': "A:1",
```
johandahlberg: Don't we have everything we need to get rid of this now?

src/sentry/static/sentry/app/views/substances/substances.jsx
```diff
@@ -36,7 +45,7 @@ class Substances extends React.Component {
       {
         Header: 'Container',
         id: 'container',
-        accessor: d => d.properties.container,
+        accessor: d => d.position.container.name,
```
johandahlberg: I guess this is related to the hard coded position property from the substance endpoint? I'm wondering when we move on with this if this should be property like it was before, or if it should be tagged directly onto the substance object?

src/sentry/static/sentry/app/components/activity/feed.jsx
```diff
@@ -40,7 +40,7 @@ const ActivityFeed = createReactClass({
     };
   },
 
-  componentWillMount() {
+  UNSAFE_componentWillMount() {
```
johandahlberg, line 43: Do we need a ticket to fix this later, since if I understand this correctly, this is deprecated?

src/sentry/static/sentry/app/redux/actions/savedSearch.js
```diff
@@ -0,0 +1,65 @@
...
+
+export const savedSearchesGet = () => dispatch => {
+  dispatch(savedSearchesGetRequest());
+  return axios
+    .get('/api/0/organizations/lab/saved-searches/')
+    .then(res => dispatch(savedSearchesGetSuccess(res.data)))
+    .catch(err => dispatch(savedSearchesGetFailure(err)));
+};
+
+// from savedSearchSelector.jsx
```
johandahlberg: Perhaps add a TODO here that this needs to be implemented later?

johandahlberg (commented):
Looks good. I'd mostly want to see some clearer comments in the code on what is done and what is not.

src/clims/api/endpoints/container.py
withrocks: Right! Need to add a checker for those tags of mine. Will fix.

src/clims/services/container.py
withrocks: Will do!

tests/clims/models/test_container.py
withrocks: Yes, I'll throw it into the core fixtures.

tests/clims/models/test_container.py
withrocks: Yeah, this is a WIP test, I'll add a skip.

src/clims/api/endpoints/substance.py
withrocks: Yeah, I'll remove the rest.

src/sentry/static/sentry/app/views/substances/substances.jsx
withrocks: I think that we had it in the properties as a hack to show it. I would prefer that we would have it on the object.

src/sentry/static/sentry/app/components/activity/feed.jsx
withrocks: Yes, I've added one.

src/sentry/static/sentry/app/redux/actions/savedSearch.js
withrocks: Right.

tests/clims/models/test_container.py
withrocks: Yes, skipping now as it's wip.

withrocks:
Thanks for the comments, I've made some fixups and a whole new commit 958bf89a. Please merge when you want. Note that there are js issues to resolve, I'll follow up with another PR for that.

PR #83, opened 2019-11-13
Introducing PluginError

johandahlberg:
`PluginError` is a mechanism for plugins to signal failures. Later on this might need to be more detailed, but for now I think this should be fine.


withrocks (approved):

withrocks:
Looks good!

PR #87, opened 2019-11-14
We need to know if excel files include formulas

johandahlberg:
To check the type of a cell (to see if it is a formula) the workbook needs to be loaded with `data_only=False`.


johandahlberg:
At this point this breaks the `TestGemstoneSampleSubmission.test_gemstone_submission_handler__with_xlsx__3_samples_found` test, if someone wants to check that out feel free to do so. Otherwise I'll look into it next week.

PR #88, opened 2019-11-15
Fix sample submission test for gemstones

GitEdvard:
I did go the easy way and removed calculation from submission file


johandahlberg:
Looks good. I'll merge this now. I think that later on we might want to move some of the excel reading from the snpseq-plugin to the core. But for now I think the way it is is alright.

PR #89, opened 2019-11-21
Improve the substance view and assign to workflow

withrocks:
Includes various changes which improve the substance view, in particular preparing it for grouping while also paging.

Please check out each individual commit for details.

Some of these are still WIP, I've added tickets for those.


src/sentry/static/sentry/app/components/substances/uploadSamples.jsx
```diff
@@ -0,0 +1,192 @@
...
+          error: err => {
+            let errors = err.responseJSON || true;
+            errors = errors.detail || true;
+            this.setState({
+              state: FormState.ERROR,
+              errors,
+            });
+          },
+          complete: () => {
+            IndicatorStore.remove(loadingIndicator);
```
johandahlberg: The loading indicator does not disappear after the samples have been uploaded, so I suspect that something is not working as expected here.

src/sentry/static/sentry/app/components/substances/uploadSamples.jsx
```diff
@@ -0,0 +1,192 @@
...
+
+  onDefaultChange(e) {
+    this.onFieldChange('isDefault', e.target.checked);
+  },
+
+  onUserDefaultChange(e) {
+    this.onFieldChange('isUserDefault', e.target.checked);
+  },
+
+  handleSubmitSucces(user) {
```
johandahlberg: This appears to be unused.

src/sentry/static/sentry/app/components/substances/workflowFilter.jsx
```diff
@@ -0,0 +1,153 @@
...
+  //   options={
+  //     tag.predefined
+  //       ? tag.values &&
+  //         tag.values.map(value => ({
+  //           value,
+  //           label: value,
+  //         }))
+  //       : this.state.options
+  //   }
+  // />
```
johandahlberg: Remove unused SelectControl?

johandahlberg (approved):
This looks good. I had some minor comments in the code itself. Poking around with this I also noticed the following bugs:
 - After a substance file has been imported the saving indicator keeps does not go away
 - Trying to change a parameter in the assign workflow modal resets the whole view
 - It is not possible to change a parameter in the assign to workflow view
 - We have no pagination of the samples page - perhaps it is preferable to show all samples matching the query rather than on showing a small subset of them? Don't know if this will cause a huge performance hit now though.
 - When opening the default `/lab` view (the first view after opening the system), nothing loads. Perhaps we want to redirect that to the `/lab/substances/` view for now?

src/sentry/static/sentry/app/components/substances/uploadSamples.jsx
withrocks: Right, I'm not sure I changed anything here, but could be some interaction between components. It can be fixed later.

src/sentry/static/sentry/app/components/substances/workflowFilter.jsx
withrocks: This file was removed in a later commit.

withrocks:
Thanks!

Regarding those bugs, (some are not bugs, i.e. they are just not implemented yet, and will be in those tickets I mentioned):

Added a ticket now:
* After a substance file has been imported the saving indicator keeps does not go away
* We have no pagination of the samples page - perhaps it is preferable to show all samples matching the query rather than on showing a small subset of them? Don't know if this will cause a huge performance hit now though.

Already in CLIMS-295
* Trying to change a parameter in the assign workflow modal resets the whole view
* It is not possible to change a parameter in the assign to workflow view

Sounds good, fixing now!
* When opening the default /lab view (the first view after opening the system), nothing loads. Perhaps we want to redirect that to the /lab/substances/ view for now?

PR #90, opened 2019-11-22
clims-289, Fix MultiFormatFile taking path as input

GitEdvard:
A similar class was defined within the sample sub validation module. I
wanted a more central place for it, since it now was used at two
locations.

### DIFF
--- a/src/clims/models/file.py
+++ b/src/clims/models/file.py
@@ -1,12 +1,44 @@
 from __future__ import absolute_import
 import os
+from abc import abstractmethod
 from django.db import models
 from tempfile import NamedTemporaryFile
 from sentry.db.models import FlexibleForeignKey, Model, sane_repr
 from openpyxl import load_workbook
 
 
-class MultiFormatFile(object):
+class MultiFormatFileBase(object):
+
+    @abstractmethod
+    def as_excel(self, read_only):
+        pass
+
+    @abstractmethod
+    def as_xml(self):
+        pass
+
+    @abstractmethod
+    def as_csv(self):
+        pass
+
+
+class MultiFormatFileFromPath(MultiFormatFileBase):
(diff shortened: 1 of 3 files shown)


johandahlberg:
I'm not sure we actually need two classes for this. Wouldn't it be better to do a single class, and then have several factory methods to initiate it? This could either take the form of a `MultiFormatFileFactor` class, or which I think is a little bit more pythonic, just two functions, one `from_organization_file(...)` and one `from_path(...)`. What do you think?

GitEdvard:
I have now added a wrapper class for MultiFormatFile, having factory methods from_path and from_organziation_file. I'm not really sure this is what you had in mind though. I was reluctant to merge it in one single class, because it would require an if-else clause in each method, and I didn't like that. I may still change my mind if you insist, so please speak your opinion!

GitEdvard:
Now, there is a refactored version of the MultiFormatFile. I have tried to avoid having multiple calls to xl. open_workbook spread out over two classes. Instead I have created internal classes of type FileContext, which deal with the temp file for organization files.

Note that tests in the snpseq repo will fail until the corresponding PR there will be merged.

src/clims/models/file.py
```diff
@@ -1,53 +1,129 @@
...
-                f.write(chunk)
+    @classmethod
+    def from_path(cls, file_path):
+        file_context = HarddiskFileContext(file_path)
+        return MultiFormatFile(file_context=file_context)
+
+    @classmethod
+    def from_organization_file(cls, organization_file):
+        file_context = OrganizationFileContext(organization_file)
+        return MultiFormatFile(file_context=file_context)
```
johandahlberg: I think these two should be made `staticmethod` rather than `classmethod`.

johandahlberg (changes requested):

PR #92, opened 2019-11-25
Clims 275 see project view

johandahlberg:
This is a first, simple implementation of the project view. The data is hard coded, and there are no tests yet, but it is a start.

### DIFF
--- a/src/sentry/static/sentry/app/components/listView.jsx
+++ b/src/sentry/static/sentry/app/components/listView.jsx
@@ -2,7 +2,6 @@ import React from 'react';
 import styled from 'react-emotion';
 import LoadingIndicator from 'app/components/loadingIndicator';
 import LoadingError from 'app/components/loadingError';
-import ListActionBar from 'app/components/listActionBar';
 import {Panel, PanelBody} from 'app/components/panels';
 import PropTypes from 'prop-types';
 import Checkbox from 'app/components/checkbox';
@@ -57,7 +56,6 @@ class ListView extends React.Component {
     columns: PropTypes.array.isRequired,
     errorMessage: PropTypes.string,
     loading: PropTypes.bool.isRequired,
-    orgId: PropTypes.string.isRequired,
     canSelect: PropTypes.bool.isRequired,
     allVisibleSelected: PropTypes.bool.isRequired,
     toggleAll: PropTypes.func.isRequired,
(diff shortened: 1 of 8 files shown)


src/sentry/static/sentry/app/components/listView.jsx
```diff
@@ -57,14 +56,14 @@ class ListView extends React.Component {
     columns: PropTypes.array.isRequired,
     errorMessage: PropTypes.string,
     loading: PropTypes.bool.isRequired,
-    orgId: PropTypes.string.isRequired,
     canSelect: PropTypes.bool.isRequired,
     allVisibleSelected: PropTypes.bool.isRequired,
     toggleAll: PropTypes.func.isRequired,
     toggleSingle: PropTypes.func.isRequired,
-    visibleIds: PropTypes.array.isRequired,
```
withrocks, line 65: I think we should be able to require this as the initial state defines it (or it should at least) as an empty array, same goes with selectedIds. That would make the code below simpler.

src/sentry/static/sentry/app/redux/reducers/projectSearchEntry.js
```diff
@@ -0,0 +1,48 @@
...
+
+// NOTE: We export the initial state in order to use it in tests
+export const initialState = {
+  loading: false,
+  errorMessage: null,
+  projectSearchEntries: [],
+  allVisibleSelected: false,
+  groupBy: 'name',
+  query: 'name',
+  byIds: {}, // The actual substance entries (TODO: have parentById and childById?)
```
withrocks, line 15: Here I would have `visibleIds` and `selectedIds` like it is in `app/redux/reducers/substanceSearchEntry.js` then you should be able to simplify the code mentioned earlier.

withrocks: ... I guess you removed it since at the moment it's not being used, but I would suggest we keep it and just fill the visible array in the reducer, like in the substance reducer. We will need paging in this view anyways, and probably selection too.

src/sentry/static/sentry/app/views/projects/projects.jsx
```diff
@@ -0,0 +1,97 @@
...
+  query: PropTypes.string,
+};
+
+const mapStateToProps = state => {
+  return {
+    projectSearchEntry: state.projectSearchEntry,
+  };
+};
+
+// TODO: Rename all functions in `mapDispatchToProps` in other files so that they match the action
```
withrocks, line 91: This TODO can be removed since it's a copy of the other one.

withrocks (changes requested):
Looks good to me! I'd suggest we'll keep the selected/visible flags as they simplify the current code and will be needed later.

withrocks:
Merging this now. Fixes will come later.

PR #95, opened 2019-11-27
clims 285, trigger sample submission validation from UI

GitEdvard:
Purpose:
Trigger sample submission validation from commonlims. 

Implementation: 
The new base class SubstancesValidationHandler is marked as an optional handler, and is called within the substances.load_file() method. Extend MultiFormatFile so that it can take a file stream as input. This is needed, since the file contents is provided by a http request (file path from client is of no use), and OrganizationFile saves contents to db, which we don't want before the validation has been performed.

### DIFF
--- a/src/clims/services/substance.py
+++ b/src/clims/services/substance.py
@@ -369,16 +369,19 @@ class SubstanceService(WrapperMixin, ExtensibleServiceAPIMixin, object):
 
         context = HandlerContext(organization=organization)
         with MultiFormatFile.from_file_stream(full_path, file_stream) as wrapped_stream:
-            plugins.handle(SubstancesValidationHandler, context, False, wrapped_stream)
+            plugins.handle(
+                SubstancesValidationHandler, context, required=False,
+                multi_format_file=wrapped_stream)
 
         logger = logging.getLogger('clims.files')
         logger.info('substance_batch_import.start')
 
         org_file = self._create_organization_file(file_stream, full_path, organization.id, logger)
 
-        # Call handler synchronously and in the same DB transaction
         with MultiFormatFile.from_organization_file(org_file) as wrapped_org_file:
-            plugins.handle(SubstancesSubmissionHandler, context, True, wrapped_org_file)
+            plugins.handle(
+                SubstancesSubmissionHandler, context, required=True,
+                multi_format_file=wrapped_org_file)
 
         return org_file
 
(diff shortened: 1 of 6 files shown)


src/clims/services/substance.py
```diff
@@ -365,11 +367,25 @@ def load_file(self, organization, full_path, fileobj, add_timestamp=False):
...
+        context = HandlerContext(organization=organization)
+        with MultiFormatFile.from_file_stream(full_path, file_stream) as wrapped_stream:
+            plugins.handle(SubstancesValidationHandler, context, False, wrapped_stream)
+
         logger = logging.getLogger('clims.files')
         logger.info('substance_batch_import.start')
 
+        org_file = self._create_organization_file(file_stream, full_path, organization.id, logger)
+
+        # Call handler synchronously and in the same DB transaction
```
johandahlberg: Not sure about this comment. If I understand this correctly these will not be called in the same transaction?

src/clims/services/substance.py
```diff
@@ -365,11 +367,25 @@ def load_file(self, organization, full_path, fileobj, add_timestamp=False):
             full_path, ext = full_path.rsplit(".", 1)
             full_path = full_path + datetime.now().strftime("%m_%d_%Y%_H_%M_%S") + "." + ext
 
-        from sentry.plugins import plugins
-        plugins.require_handler(SubstancesSubmissionHandler)
+        context = HandlerContext(organization=organization)
+        with MultiFormatFile.from_file_stream(full_path, file_stream) as wrapped_stream:
+            plugins.handle(SubstancesValidationHandler, context, False, wrapped_stream)
```
johandahlberg: I think it could be useful for both the handle methods here to assign the parameters with keywords. Right now it is difficult to understand especially what `False` and `True` is actually setting.

johandahlberg (commented):
This looks good I think. I had one question and a suggestion for a minor fix. Perhaps we want to add tests for this new methods, further down the line as well?

src/clims/services/substance.py
GitEdvard: This comment was here before. Probably, it looks like a new comment because I have moved some code around. Should I remove the comment?

src/clims/services/substance.py
GitEdvard: Oh, actually I had the same thought. I'll look into it. The reason for I didn't do anything was, on a deeper level there is an \*args in the signature after the True/False arg.

src/clims/services/substance.py
johandahlberg: @withrocks This sound like a TODO thing? Or how does this actually work?

src/clims/services/substance.py
johandahlberg: That shouldn't be a problem I think, as long as all the other parameters are given as kwargs.

GitEdvard:
Ok, I think I'm ready with this!

PR #97, opened 2019-11-28
clims 292, parse Excel cell values from internal clims ExcelFile

GitEdvard:
Purpose:
1) Make the code for parsing Excel cell values available for re-use in commonlims framework
2) Alert external developers that it's a problem with calculated fields in Excel

### DIFF
--- a/src/clims/models/file.py
+++ b/src/clims/models/file.py
@@ -5,7 +5,7 @@ from django.db import models
 from tempfile import NamedTemporaryFile
 from sentry.db.models import FlexibleForeignKey, Model, sane_repr
 from openpyxl import load_workbook
-from six import binary_type
+from six import text_type
 from six import string_types
 
 
@@ -202,7 +202,7 @@ class ClimsExcelFile:
         elif isinstance(cell.value, string_types):
             return cell.value.encode('utf-8')
         else:
-            return binary_type(cell.value)
+            return text_type(cell.value)
 
 
 class OrganizationFile(Model):
--- a/src/clims/models/file.py
+++ b/src/clims/models/file.py
@@ -204,6 +204,9 @@ class ClimsExcelFile:
         else:
             return text_type(cell.value)
 
+    def save(self, filename):
+        self._workbook.save(filename)
+
 
 class OrganizationFile(Model):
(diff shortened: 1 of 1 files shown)


src/clims/models/file.py
```diff
@@ -175,6 +177,34 @@ def _validate_file_path(self):
...
+            raise ValueError('Excel calculated values are not supported! openpyxl uses cached '
+                             'values for calculated fields, which are populated every last '
+                             'time the doc is opened by MicroSoft Excel software. There is '
+                             'currently no open-source software that evaluates Excel formulas')
+        elif cell.value is None:
+            return None
+        elif isinstance(cell.value, string_types):
+            return cell.value.encode('utf-8')
+        else:
+            return binary_type(cell.value)
```
johandahlberg: Does this ever happen? I am hard pressed to think of a case where the data returned would not be possible to encode as unicode. Do you have any example of this? Otherwise perhaps we'd want to raise an exception here instead?

johandahlberg (approved):
Looks cool. I had one questions. I'm also thinking that perhaps we should move the conversions in here, so that you could do something like `read_int_from_cell`. But we can do that later.

src/clims/models/file.py
GitEdvard: I'm not really sure what you mean here

GitEdvard: regarding cell.value.encode(), I have had problems importing string values, although I don't remember the details unfortunately. encode('utf-8') solved it.

GitEdvard: regarding cell.binary_type(), as I recall, all int and floats goes in this. when I did "normal" conversion to string I got a message from flake8/pylint (?) to use binary_type instead, which will work in python3 as well.

johandahlberg: I'm not 100% sure about how oppyxls works here, but if you look at https://six.readthedocs.io/#six.binary_type I think this might cause us some trouble when we switch to Python 3, since I think that the rest of the code expects a string-like type?

GitEdvard: I've changed it to text_type instead.

PR #98, opened 2019-11-28
Implement workflow assignment UI

withrocks:
Fixes issues with the workflow assignment view. Note that the backend is not in scope here, so this is still using a mocked backend response.


johandahlberg:
LGTM

PR #100, opened 2019-12-03
Clims 276 search for projects

johandahlberg:
This is a first version of the projects endpoint. It has the same "basic search" as the substance view. I also moved the search into the extensible service API to avoid duplicating the search code all over the place. However, I do imagine that in the relatively near future we will want to move all the search code into it's own search engine proxy class, or similar.

This new search endpoint has not yet been hooked up to the front-end. That will need to happen in a separate PR.

### DIFF
--- a/src/sentry/testutils/fixtures.py
+++ b/src/sentry/testutils/fixtures.py
@@ -752,7 +752,6 @@ class Fixtures(object):
         return ret
 
     def create_substance(self, klass, name=None, **kwargs):
-        # TODO: This shouldn't be necessary as there should be only one app instance per test
         properties = kwargs or dict()
         ret = self.register_extensible(klass)
 
@@ -762,7 +761,6 @@ class Fixtures(object):
         return ret
 
     def create_clims_project(self, klass, name=None, **kwargs):
-        # TODO: This shouldn't be necessary as there should be only one app instance per test
         properties = kwargs or dict()
         ret = self.register_extensible(klass)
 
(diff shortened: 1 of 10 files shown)


src/sentry/testutils/fixtures.py
```diff
@@ -761,6 +761,16 @@ def create_substance(self, klass, name=None, **kwargs):
         ret = self.app.extensibles.create(name, klass, self.organization, properties=properties)
         return ret
 
+    def create_clims_project(self, klass, name=None, **kwargs):
+        # TODO: This shouldn't be necessary as there should be only one app instance per test
```
withrocks: I think this is an old comment that I left that didn't make sense anymore.

tests/clims/api/endpoints/test_projects.py
```diff
@@ -0,0 +1,93 @@
...
+
+from sentry.testutils import APITestCase
+
+from rest_framework import status
+from tests.fixtures.plugins.gemstones_inc.models import GemstoneProject
+
+
+class ProjectTest(APITestCase):
+    def create_gemstone(self, *args, **kwargs):
+        return self.create_Project(GemstoneProject, *args, **kwargs)
```
withrocks: Should probably be `create_clims_project`?

withrocks: I don't think this method is being used below.

withrocks (commented):
Looks good! Two minor comments.

PR #102, opened 2019-12-04
Clims 276 hook up project view to backend

johandahlberg:
This hooks up the project view to the backend, including enabling pagination.

It also adds some other minor fixes (in separate commits):
 - allows prop names to be set explicitly if the users wants
 - has the data generation script create projects

### DIFF
--- a/src/clims/runner/commands/create_example_data.py
+++ b/src/clims/runner/commands/create_example_data.py
@@ -61,4 +61,4 @@ def createexampledata():
         name = "demo-{}".format(uuid4().hex)
         project = ExampleProject(name=name, organization=org, project_code=name, pi=random.choice(pis))
         project.save()
-        click.echo("Created sample: {}".format(project.name))
+        click.echo("Created project: {}".format(project.name))
--- a/src/clims/runner/commands/create_example_data.py
+++ b/src/clims/runner/commands/create_example_data.py
@@ -1,6 +1,8 @@
 
 from __future__ import absolute_import, print_function
 import random
+from uuid import uuid4
+
 
 import click
 from sentry.runner.decorators import configuration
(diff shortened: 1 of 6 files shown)


src/sentry/static/sentry/app/views/projects/projects.jsx
```diff
@@ -54,19 +96,27 @@ class Projects extends React.Component {
...
+            query={query}
           />
           <ListView
             orgId={this.props.organization.id}
             columns={this.getHeaders()}
-            dataById={this.props.projectSearchEntry.byIds}
+            dataById={byIds}
             canSelect={false}
-            visibleIds={Object.keys(this.props.projectSearchEntry.byIds)}
+            visibleIds={Object.keys(byIds)}
```
withrocks: There would be an issue with this later on if we stop clearing the `byIds` dictionary on every fetch, e.g. if we want to keep the prev. and next page in memory. I would just use `this.props.projectSearchEntry.visibleIds` and set that in the reducer, so the UI doesn't need to worry about it.

But I don't think it needs to be fixed now.

withrocks (approved):
Looks good! One minor comment, can merge if you want me to.

src/clims/runner/commands/create_example_data.py
```diff
@@ -47,3 +55,10 @@ class ExampleSample(SubstanceBase):
                                erudite=random.randint(1, 100))
         sample.save()
         click.echo("Created sample: {}".format(sample.name))
+
+    pis = ["Rosaline Franklin", "Charles Darwin", "Gregor Medel"]
+    for _ in range(100):
+        name = "demo-{}".format(uuid4().hex)
+        project = ExampleProject(name=name, organization=org, project_code=name, pi=random.choice(pis))
+        project.save()
+        click.echo("Created sample: {}".format(project.name))
```
withrocks: `s/sample/project`

johandahlberg:
This is fixed now.

PR #103, opened 2019-12-05
When value have not been prefetched, go get them

johandahlberg:
I'm not sure this is optimal, since for some cases it could lead to unexpected db calls. I.e. you expected to have all the properties prefetched, but in reality they are not, and every get will generate a call to the database.

This is a good enough sketch for now though, I suppose? Later we might want to prefetch them and add them to the object if they are not there. What do you think?


withrocks (approved):
I agree that this is good for now. I believe that we (at least I) want to go and give this whole system a good refactoring when it's convenient anyways, and the most important thing for now is to create passing tests that define how the system should work.

PR #108, opened 2019-12-19
Clims 278 group by sample type

johandahlberg:
This should enable getting values, either all or all unique, values for a property for all extensible types. Right now you need to feed it the `extensible_type`, e.g. `GemstoneSample`. We might instead have the type be picked up from the context (i.e. put the type as a field on the service class), but I think that this is clearer and gives the user more control.

This means that the API needs will look something like this:
```
self.app.substances.get_unique_values_of_property(property='color', extensible_type=GemstoneSample)
```
Let me know what you think.

### DIFF
--- a/src/clims/services/extensible_service_api.py
+++ b/src/clims/services/extensible_service_api.py
@@ -105,23 +105,26 @@ class ExtensibleServiceAPIMixin(object):
                 get_args['archetype__{}'.format(key)] = value
         return get_args
 
-    def _get_values_qs(self, property, extensible_type):
+    @staticmethod
+    def _get_values_qs(property, extensible_type):
         extensible_type_name = extensible_type.type_full_name_cls()
         property_query_set = ExtensibleProperty.objects.filter(
             extensible_property_type__extensible_type__name=extensible_type_name,
             extensible_property_type__name=property).select_related('extensible_property_type')
         return property_query_set
 
-    def get_values_of_property(self, property, extensible_type):
-        property_query_set = self._get_values_qs(property, extensible_type)
+    @staticmethod
+    def get_values_of_property(property, extensible_type):
+        property_query_set = ExtensibleServiceAPIMixin._get_values_qs(property, extensible_type)
         extensible_type = property_query_set[0].extensible_property_type
         extensible_type_value_field = extensible_type.get_value_field()
         values = property_query_set.all().values(extensible_type_value_field)
         return ([x[extensible_type_value_field] for x in values])
 
-    def get_unique_values_of_property(self, property, extensible_type):
+    @staticmethod
+    def get_unique_values_of_property(property, extensible_type):
         # TODO: add organization to the filter parameters
-        property_query_set = self._get_values_qs(property, extensible_type)
+        property_query_set = ExtensibleServiceAPIMixin._get_values_qs(property, extensible_type)
         extensible_type = property_query_set[0].extensible_property_type
         extensible_type_value_field = extensible_type.get_value_field()
         unique_values = property_query_set.order_by(extensible_type_value_field).distinct(extensible_type_value_field)
--- a/src/clims/services/extensible_service_api.py
+++ b/src/clims/services/extensible_service_api.py
@@ -116,16 +116,26 @@ class ExtensibleServiceAPIMixin(object):
     @staticmethod
     def get_values_of_property(property, extensible_type):
         property_query_set = ExtensibleServiceAPIMixin._get_values_qs(property, extensible_type)
-        extensible_type = property_query_set[0].extensible_property_type
-        extensible_type_value_field = extensible_type.get_value_field()
-        values = property_query_set.all().values(extensible_type_value_field)
-        return ([x[extensible_type_value_field] for x in values])
+        try:
+            extensible_type = property_query_set[0].extensible_property_type
+            extensible_type_value_field = extensible_type.get_value_field()
+            values = property_query_set.all().values(extensible_type_value_field)
+            return ([x[extensible_type_value_field] for x in values])
+        except IndexError:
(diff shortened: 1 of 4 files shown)


src/clims/services/extensible_service_api.py
```diff
@@ -103,3 +104,25 @@ def _get_filter_arguments(self, **kwargs):
             else:
                 get_args['archetype__{}'.format(key)] = value
         return get_args
+
+    def _get_values_qs(self, property, extensible_type):
+        extensible_type_name = extensible_type.type_full_name_cls()
+        property_query_set = ExtensibleProperty.objects.filter(
+            extensible_property_type__extensible_type__name=extensible_type_name,
+            extensible_property_type__name=property).select_related('extensible_property_type')
```
GitEdvard, line 113: I'm curious, why use select_related instead of prefetch_related?

src/clims/services/extensible_service_api.py
```diff
@@ -103,3 +104,25 @@ def _get_filter_arguments(self, **kwargs):
...
+        return property_query_set
+
+    def get_values_of_property(self, property, extensible_type):
+        property_query_set = self._get_values_qs(property, extensible_type)
+        extensible_type = property_query_set[0].extensible_property_type
```
GitEdvard: Perhaps this is wrong spelling? extenstible_type already means something and it's not extensible_property_type.

GitEdvard (commented):
This look great, although I have a few comments. In cases when there is multiple choice fields (enum fields), my idea was that these values should be put into a separate table, and probably, the search in this PR would be toward that table.

This seem to me like one step closer for solving task 278. Is the thing with grouping samples over a selected property solved already? (I don't see that part in this PR)

johandahlberg:
Yeah - it is only going part of the way towards solving clims-278. But I wanted to break it out into it's own PR since it was required to do it, but had wider implication for the code as a whole.

src/clims/services/extensible_service_api.py
johandahlberg: They seem to be rather similar: https://stackoverflow.com/questions/31237042/whats-the-difference-between-select-related-and-prefetch-related-in-django-orm Not actually sure which would be best to use in this scenario - probably something that we could get to looking at in a profiler later on.

withrocks: I think your usage here makes sense, because it translates into a natural SQL join.

src/clims/services/extensible_service_api.py
```diff
@@ -103,3 +104,38 @@ def _get_filter_arguments(self, **kwargs):
...
+        property_query_set = ExtensibleProperty.objects.filter(
+            extensible_property_type__extensible_type__name=extensible_type_name,
+            extensible_property_type__name=property).select_related('extensible_property_type')
+        return property_query_set
+
+    @staticmethod
+    def get_values_of_property(property, extensible_type):
+        property_query_set = ExtensibleServiceAPIMixin._get_values_qs(property, extensible_type)
+        try:
+            extensible_type = property_query_set[0].extensible_property_type
```
withrocks: I would use `get` rather than `[0]`. It returns an error if there are more than one returned too as well as `DoesNotExist` in the case you're catching.

https://docs.djangoproject.com/en/1.9/ref/models/querysets/#get

src/clims/services/extensible_service_api.py
```diff
@@ -103,3 +104,38 @@ def _get_filter_arguments(self, **kwargs):
...
+        except IndexError:
+            raise DoesNotExist("No property with name: {} found on extensible type: {}".format(property,
+                                                                                               extensible_type))
+
+    @staticmethod
+    def get_unique_values_of_property(property, extensible_type):
+        # TODO: add organization to the filter parameters
+        try:
+            property_query_set = ExtensibleServiceAPIMixin._get_values_qs(property, extensible_type)
+            extensible_type = property_query_set[0].extensible_property_type
```
withrocks: Same here (`get`)

src/clims/services/extensible_service_api.py
```diff
@@ -103,3 +104,38 @@ def _get_filter_arguments(self, **kwargs):
...
+            extensible_type_value_field = extensible_type.get_value_field()
+            unique_values = property_query_set\
+                .order_by(extensible_type_value_field)\
+                .distinct(extensible_type_value_field)
+            return set([x.value for x in unique_values])
```
withrocks: Just reminder: One can also return `{x.value for x in unique_values}` directly (set comprehension).

src/clims/services/extensible_service_api.py
```diff
@@ -103,3 +104,38 @@ def _get_filter_arguments(self, **kwargs):
...
+
+    @staticmethod
+    def get_values_of_property(property, extensible_type):
+        property_query_set = ExtensibleServiceAPIMixin._get_values_qs(property, extensible_type)
+        try:
+            extensible_type = property_query_set[0].extensible_property_type
+            extensible_type_value_field = extensible_type.get_value_field()
+            values = property_query_set.all().values(extensible_type_value_field)
+            return ([x[extensible_type_value_field] for x in values])
+        except IndexError:
```
withrocks, line 126: If you change to not using an index, you would be able to catch `DoesNotExist` instead.

tests/clims/services/test_projects.py
```diff
@@ -1,3 +1,41 @@
...
+import random
+
+
+class TestProjectService(TestCase):
+
+    def setUp(self):
+        self.register_extensible(GemstoneProject)
+
+    def create_a_bunch_of_projects(self):
+        country_choices = ['Sweden', 'Norway', 'Denmark', 'Finland', 'Flat Iceland']
```
withrocks, line 15: Or `Flatland` :D

PR #109, opened 2019-12-19
clims 277, group samples by container, fix container endpoint

GitEdvard:
Purpose:
Prepare the container endpoint to be able to show an expandable list of containers in UI. User first see a list of containers. When user clicks on a selected container, all samples in that container are shown. 

Implementation:
Fix an expanded http response from the container endpoint, that contains json for all contained samples. 

Refactoring:
I intended to remove low level django quering, and container wrapping from the container endpoint. For this I created the ContainerQueryBuilder. I also called the base class .filter method in order to do the wrapping.

### DIFF
--- a/src/clims/api/serializers/models/substance.py
+++ b/src/clims/api/serializers/models/substance.py
@@ -4,6 +4,14 @@ from clims.api.serializers.models.extensible_property import ExtensiblePropertyS
 
 from rest_framework import serializers
 from rest_framework.fields import DictField
+from six import text_type
+
+
+class ContainerIndexField(serializers.Field):
+    def to_representation(self, value):
+        return {
+            "index": text_type(value)
+        }
 
 
 class LocationField(serializers.Field):
@@ -26,3 +34,4 @@ class SubstanceSerializer(serializers.Serializer):
     properties = DictField(child=ExtensiblePropertySerializer(read_only=True))
     type_full_name = serializers.CharField()
     location = LocationField(read_only=True)
+    container_index = ContainerIndexField(read_only=True)
--- a/src/clims/services/container.py
+++ b/src/clims/services/container.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 
 import re
 import six
+from six import iteritems
 from clims.services.extensible import ExtensibleBase
 from clims.models import Container, ContainerVersion
 from clims.services.wrapper import WrapperMixin
@@ -157,6 +158,14 @@ class ContainerBase(ExtensibleBase):
         # Update the value. This will not actually move it in the backend until either
         # the container is saved
         self._locatables[ix.raw] = value
+
+    @property
+    def contents(self):
+        c = []
+        for loc_raw, substance in iteritems(self._locatables):
+            c.append(substance)
+
+        return c
 
     def __getitem__(self, key):
         ix = self.IndexType.from_any_type(key)
--- a/src/clims/services/container.py
+++ b/src/clims/services/container.py
@@ -262,41 +262,36 @@ class ContainerService(WrapperMixin, ExtensibleServiceAPIMixin, object):
         return self.get(name=name)
 
     def filter_from(self, query_builder):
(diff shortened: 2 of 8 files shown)


src/clims/api/serializers/models/substance.py
```diff
@@ -4,6 +4,14 @@
...
 from rest_framework import serializers
 from rest_framework.fields import DictField
+from six import text_type
+
+
+class ContainerIndexField(serializers.Field):
+    def to_representation(self, value):
+        return {
+            "index": text_type(value)
+        }
```
johandahlberg, line 14: I'm not sure I understand this - but isn't this the same thing as `LocationField`? Or at least that the intent of both are pretty much the same?

src/clims/services/container.py
```diff
@@ -158,6 +159,14 @@ def __setitem__(self, key, value):
         # the container is saved
         self._locatables[ix.raw] = value
 
+    @property
+    def contents(self):
+        c = []
```
withrocks, line 164: FYI: I've also added a `contents` attribute to the container - but it's in my next PR. I had it return only the items that have something in it, while iterating through the container itself returns all items (even the empty ones). I'll synch with you when I merge the two.

src/clims/services/container.py
```diff
@@ -252,22 +261,37 @@ def __init__(self, app):
...
 
-    def _search_qs(self, query):
-        # TODO: This is temporary. We will be using elastic for searching.
-        query = query.strip()
+    def filter_from(self, query_builder):
+        query_params = query_builder.parse_query()
+        return self.filter(**query_params)
+
+
+class ContainerQueryBuilder:
```
withrocks, line 269: I like that we have a class similar to this. It will encapsulate Elastic stuff later.

src/clims/services/container.py
GitEdvard: Ok!

src/clims/services/container.py
GitEdvard: :)

src/clims/api/serializers/models/substance.py
GitEdvard: container_index sits on SubstacneBase, and location is a raw django model. From SubstanceBase, I have access to the container and container type the substance is located in, and from container_type I can get the plate-index representation of a location ("A:1"). From within SubstanceLocation (raw django model) I do not have access to anything. The index  representation from location is of the form "(0,1,0)"

GitEdvard: I know that the current state is confusing, there is index in container_index, and index is also within location. My idea is that the location property should be remade.

GitEdvard: See also this TODO: https://github.com/commonlims/commonlims/blob/develop/src/clims/services/substance.py#L189

src/clims/services/substance.py
```diff
@@ -180,6 +180,18 @@ def project(self):
...
 
+    @property
+    def container_index(self):
+        # TODO: the container should be in a local cache!
+        # when looping samples in a container, the very same container is fetched over and
+        # over again
+        loc = self.location
+        container = self._app.containers.to_wrapper(loc.container)
+        container_index_class = container.IndexType
+        # Can we assume that container index class has only row and column, not z?
```
withrocks, line 191: No, the idea was that it should always have the three, and I'm getting an error on it now with older data. I've got a suggestion of how to solve it which I'll commit to this PR. You can then rebase if you want to. I'll ping you when it's ready.

withrocks: Also, as @johandahlberg pointed out, this has the same role as the location property, but that's now returning these raw indexes. That one should be updated instead I think. I'll add that to my commit.

withrocks: And regarding the performance, I agree that we'll have to find a clean solution to that. Prefetching comes first to my mind - again.

GitEdvard: Ok

PR #110, opened 2020-01-07
Use docker for all middleware

withrocks:


Makefile
```diff
@@ -11,54 +11,24 @@ PIP = LDFLAGS="$(LDFLAGS)" pip
...
+middleware-teardown:
+	@./scripts/stop-local-middleware.sh
 
 create-example-data:
 	@echo "--> Add example data"
 	lims createexampledata
 
-
-# TODO Don't know if we should create the user here, since it is not super intuative to reset the
-#      db to create a user. /JD 2019-10-03
```
withrocks, line 60: Removed this comment as it's now documented that for a fresh dev system one uses `make fresh`, which I guess must include having a test user. Since this is the first step everyone goes through it makes sense to me to have it there. Tell me what you think.

PR #112, opened 2020-01-07
Show rounded value of volumes in sample list

GitEdvard:
Purpose:
Add function "showRounded" to show rounded values in lists. This function is located on a central place so that it may be referenced by all components.

### DIFF
--- a/src/sentry/static/sentry/app/views/substances/substances.jsx
+++ b/src/sentry/static/sentry/app/views/substances/substances.jsx
@@ -14,6 +14,7 @@ import ListView from 'app/components/listView';
 import SentryTypes from 'app/sentryTypes';
 import ListActionBar from 'app/components/listActionBar';
 import Pagination from 'app/components/pagination';
+import {showRounded} from 'app/utils/formatters';
 
 class Substances extends React.Component {
   constructor(props) {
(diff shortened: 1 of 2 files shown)


src/sentry/static/sentry/app/views/substances/substances.jsx
```diff
@@ -92,7 +93,8 @@ class Substances extends React.Component {
       {
         Header: 'Volume',
         id: 'volume',
-        accessor: d => (d.properties.volume ? d.properties.volume.value : null),
+        accessor: d =>
```
withrocks, line 96: There is support for (localized) number formatting in the `Intl.NumberFormat` object which you might want to check out. It's supported in all relevant browsers:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat
https://caniuse.com/#search=DateTimeFormat

GitEdvard:
Now I've updated to use the Intl.NumberFormat method. The parameter locale=undefined should show number format according to settings in the local machine. However, I had problem testing this (I have to restart my machine and it is cumbersome :-(  ). Maybe we can merge this and possibly change it later if needed?

withrocks (approved):

withrocks:
Sounds good. The CI build is failing now in the linter step: https://travis-ci.com/commonlims/commonlims/jobs/273920458

Are you not getting errors when you commit? If not I can come and take a look at why (should let you know if you don't commit with `--no-verify`)

withrocks:
Turned out that the error is in the develop branch. I'll rebase and push a fix to this branch.

PR #114, opened 2020-01-08
Get properties from samples and projects

johandahlberg:
There is an issue here that needs to be resolved, relating to that in the current implementation you need to know which extensible type you want to get your properties from. I think that this should be somehow
hidden from the user, but we need to figure out how to do this.

I'm submitting this PR now, so we can have a look at how to solve it together.


GitEdvard:
Interesting. If one stands in the samples page for example, I would like the search/filtering performed against the same dataset whether I have search criteria "sample.name:sample", or "sample.color:red". Today, the searches on sample name includes all sample types, (organization is not yet included either). 

I think it depends on how the user wants it here, whether she want to see all sample types that exists within the current organization, or if the search should be limited to a certain type of sample. In our case, at snpseq, I can imagine when showing samples within a working step, that there might be both rml-samples and prep-samples within the same working step. Different sample types may also share the same type of property, like volume. 

Perhaps we could first check which substance types that have the desired property, and then include all these in the search?

johandahlberg:
This has now been updated, and I have implemented methods to get properties of substances and projects. I don't know how much sense it makes for have an endpoint to get properties of containers, but if you like me to do so I can easily add such an additional endpoint.

johandahlberg:
Ok, so I talked to @withrocks and we decided we might as well do the container thing right away.

withrocks (approved):
Looks mighty fine!

PR #115, opened 2020-01-10
Simplify extensible services

withrocks:
The services for extensible objects where rather complex, partly because they were using mixins rather than composition.


src/clims/services/base_extensible_service.py
```diff
@@ -116,9 +155,9 @@ def _get_values_qs(property, extensible_type):
             extensible_property_type__name=property).select_related('extensible_property_type')
         return property_query_set
 
-    @staticmethod
-    def get_values_of_property(property, extensible_type):
-        property_query_set = ExtensibleServiceAPIMixin._get_values_qs(property, extensible_type)
+    @classmethod
```
johandahlberg, line 158: I'm not against it, but I'd be interested in why you think that a `classmethod` is more appropriate than a `staticmethod` here.

johandahlberg (approved):
LGTM.

src/clims/services/base_extensible_service.py
withrocks: That's because we're using the class in the body of the method. It's just so that it's easier to refactor later, if the class changes name.

PR #116, opened 2020-01-13
clims 320, show all samples and projects at first visit

GitEdvard:
Purpose:
When user first visits the Samples or the Projects view, show all items as default.


withrocks (approved):
So it was that simple :)

PR #118, opened 2020-01-15
Refactors and changes PluginManager, making it simpler to test and understand

withrocks:


src/clims/handlers.py
```diff
@@ -21,10 +31,154 @@ class HandlerContext(object):
...
+    def handler_count(self, cls):
+        return len(self.handlers[cls])
+
+    def require_single_handler(self, cls):
+        handlers = self.handlers[cls]
+        count = len(handlers)
+        if count == 0:
+            raise RequiredHandlerNotFound("No handler that implements '{}' found".format(cls))
+        elif count > 1:
+            # TODO: New type or change the type name
```
johandahlberg: I don't understand this comment. Is the idea that a new type should be generated here, or is it simply something to notify the user about?

src/clims/handlers.py
```diff
@@ -21,10 +31,154 @@ class HandlerContext(object):
...
+        logger.info("Handling action {}".format(cls))
+
+        ret = list()
+        if required:
+            self.require_handler(cls)
+        handlers = self.handlers[cls]
+        logger.debug("Handlers found for '{}': '{}'".format(cls, handlers))
+
+        for handler in handlers:
+            # TODO: the plugins manager should have an instance of the app
```
johandahlberg, line 99: Perhaps can go ahead and fix this now, since it should be a fairly small and simple fix?

src/clims/handlers.py
```diff
@@ -21,10 +31,154 @@ class HandlerContext(object):
...
+            ret.update(find_implementations_in(mod.__name__, handler_type))
+            for _, mod_name, _ in pkgutil.iter_modules(path=mod.__path__, prefix=mod.__name__ + "."):
+                ret.update(find_implementations_in(mod_name, handler_type))
+            return ret
+
+        for handler_type in self.handlers.keys():
+            logger.debug("Searching for implementations of {}".format(handler_type))
+            self.handlers[handler_type].update(find_all_implementations(handler_type))
+
+    def to_handler_config(self):
```
johandahlberg, line 142: Great idea!

src/clims/handlers.py
```diff
@@ -21,10 +31,154 @@ class HandlerContext(object):
...
+        """
+        Returns a YML view of the state of handlers
+        """
+        import yaml
```
johandahlberg: Move import to top?

src/clims/handlers.py
```diff
@@ -21,10 +31,154 @@ class HandlerContext(object):
...
+                ImplOfX(X)
+                ImplOfY(Y)
+            PluginB:
+                ImplOfY2(ImplOfY)
+
+        ImplOfY2 from PluginB should be used but not ImplOfY from PluginA.
+
+        This would probably be more expected than if both handlers would be used, but
+        could also lead to unexpected situations, so it's still being discussed by the team.
+        """
```
johandahlberg, line 177: I guess we need to come to some conclusion on whether we want to do this via config or not? But we don't need to address that in this PR.

src/clims/services/substance.py
```diff
@@ -403,12 +403,13 @@ def _create_organization_file(self, file_stream, full_path, organization_id):
     def create_submission_demo(self, file_type):
         import os
         from clims.handlers import SubstancesSubmissionFileDemoHandler
-        plugins.require_single_handler(SubstancesSubmissionFileDemoHandler)
-        handlers = plugins.handle(SubstancesSubmissionFileDemoHandler, None, True, file_type)
-        handler = handlers[0]
+        self._app.plugins.handlers.require_single_handler(SubstancesSubmissionFileDemoHandler)
+        handlers = self._app.plugins.handlers.handle(SubstancesSubmissionFileDemoHandler, None, True, file_type)
+        handler = utils.single(handlers)
```
johandahlberg, line 408: Nice to see this in action. Looks nice and clean.

src/sentry/plugins/base/manager.py
```diff
@@ -14,57 +14,73 @@
...
+    Plugins need to be installed via a call to `lims upgrade`. This will install all plugins
+    that are found in the application at the time.
 
-    def __len__(self):
-        return sum(1 for i in self.all())
+    When a plugin is found in the environment, it's installed, i.e. added to the database.
+    After that, it needs to exist on load time from there on. It's currently not supported to
+    uninstall a plugin.
+
+    When the application loads, it will load all plugins that exist in the database.
```
johandahlberg, line 42: I think that we should perhaps start working these things into the general documentation as well? So that users know what to expect with regards to plugin install procedures.

src/sentry/plugins/base/manager.py
```diff
@@ -75,72 +91,262 @@ def register_plugin_workflow(self, plugin):
...
+
+        for ep in entry_points:
+            try:
+                plugin = ep.load()
+                yield plugin
+            except Exception:  # Handling all exceptions since the code is unknown to us.
+                import traceback
+                click.echo(
+                    "Failed to load plugin %r:\n%s" % (ep.name, traceback.format_exc()),
+                    err=True)
```
johandahlberg, line 194: To me it feels a bit like a leaky abstraction to use `click` to return the error here. Shouldn't we be able to use normal logging, and then re-raise the exception?

src/sentry/plugins/base/manager.py
```diff
@@ -14,57 +14,73 @@
...
+
+    When the application loads, it will load all plugins that exist in the database.
+    """
+
+    def __init__(self, instance_manager):
+        self.handlers = HandlerManager()
+        self.instance_manager = instance_manager
+
+    # Install (during upgrade)
+
```
johandahlberg, line 50: I like this way of having subsections in the file for different things. Perhaps you could add a comment about that to the top of the file, so that a developer coming into it immediately know that this is the way things are organized?

src/sentry/plugins/base/manager.py
```diff
@@ -75,72 +91,262 @@ def register_plugin_workflow(self, plugin):
...
+        isn't installed anymore, or has a different version, an error is raised.
+        """
+        logger.info("Loading all installed plugins")
+        from clims.models import PluginRegistration
+        try:
+            installed = list(PluginRegistration.objects.all())
+        except ProgrammingError:
+            # We might be loading the application before migrations have run, so the
+            # PluginRegistration type doesn't exist. In this case we silently pass and no plugins
+            # will be loaded
```
johandahlberg, line 219: Don't you think that we should add a message about this to the log? Seems like a thing that could be nice to see there.

src/sentry/plugins/base/manager.py
```diff
@@ -75,72 +91,262 @@ def register_plugin_workflow(self, plugin):
...
+            # an acceptably functioning system. For now however, this error is raised
+
+            # Allow the user to ignore the plugin if an environment variable is set. This
+            # is mainly for debug purposes and to be able to run `lims shell` in this situation.
+            if not os.environ.get("CLIMS_IGNORE_UNAVAILABLE_PLUGINS", None) == "1":
+                six.reraise(RequiredPluginCannotLoad,
+                    "Can't import required plugin {}@{}. The plugin has been installed e.g. via "
+                    "`lims upgrade` but the implementation is not found in the python environment. "
+                    "environment variable CLIMS_IGNORE_UNAVAILABLE_PLUGINS=1".format(
+                        plugin_registration.name, plugin_registration.version))
```
johandahlberg, line 262: Good idea!

src/sentry/plugins/base/manager.py
```diff
@@ -75,72 +91,262 @@ def register_plugin_workflow(self, plugin):
...
+        if not mod:
+            logger.info("No handlers module found in plugin '{}'".format(plugin))
+        else:
+            logger.info("Loading all handlers in plugin '{}'".format(plugin.get_name_and_version()))
+            self.handlers.load_handlers(mod)
+
+        self.handlers.validate()
+
+    def init_plugin_instance(plugin):
+        # TODO: Call this when the plugin is run on load time
```
johandahlberg, line 281: Not sure I understand this TODO. Is this not done today?

src/sentry/plugins/base/manager.py
```diff
@@ -75,72 +91,262 @@ def register_plugin_workflow(self, plugin):
...
+
+    def get(self, slug):
+        for plugin in self.all(version=None):
+            if plugin.slug == slug:
+                return plugin
+        raise KeyError(slug)
+
+    # Legacy
+
+    # These methods are pending deletion (from the sentry core)
```
johandahlberg, line 349: I have not reviewed below this point.

src/sentry/testutils/cases.py
```diff
@@ -90,7 +87,12 @@ def setup_services(self):
         # Set up clean service classes. This should ensure that each test case has a clean cache
         # as caches should only be in services.
         ioc.set_application(ApplicationService())
+
+        # NOTE: We'll have to reset the plugin manager so we don't get the plugins
+        # that have been loaded in scope. It would be better to never load those plugins in the
+        # first place, but that will be fixed when we stop using the global `plugins` object.
```
johandahlberg, line 93: Perhaps it would be good to add some kind of cross-link between these comments, so that when we fix one, we fix the other?

src/sentry/utils/managers.py
```diff
@@ -1,70 +1,112 @@
...
-    def __init__(self, class_list=None, instances=True):
-        if class_list is None:
-            class_list = []
-        self.instances = instances
-        self.update(class_list)
+
+    ImportException = ImportException
+
+    InitializeException = InitializeException
+
```
johandahlberg, line 35: I have seen you do this, but I don't understand why? Just out of curiosity, could you explain?

src/clims/plugins/demo/handlers/create_example_data.py
```diff
@@ -0,0 +1,53 @@
...
+            sample = ExampleSample(name=name,
+                                   organization=self.context.organization,
+                                   moxy=random.randint(1, 100),
+                                   cool=random.randint(1, 100),
+                                   erudite=random.randint(1, 100),
+                                   sample_type=random.choice(available_sample_types))
+            sample.save()
+            logger.info("Created sample: {}".format(sample.name))
+
+        pis = ["Rosaline Franklin", "Charles Darwin", "Gregor Mendel"]
```
johandahlberg: There is a typo in the name here Rosaline -> Rosalind. I wrote it originally, so no big surprise there.

src/clims/runner/commands/create_example_data.py
```diff
@@ -13,55 +11,26 @@ def createexampledata():
...
-    for _ in range(100):
-        name = "demo-{}".format(uuid4().hex)
-        project = ExampleProject(name=name, organization=org, project_code=name, pi=random.choice(pis))
-        project.save()
-        click.echo("Created project: {}".format(project.name))
+    # Now we'll run the handlers for all plugins (one of them being the DemoPlugin) for
+    # creating example data. All plugins can plug into this behaviour by implementing the
+    # CreateExampleDataHandler:
+    context = HandlerContext(organization=org)
+    app.plugins.handlers.handle(CreateExampleDataHandler, context, required=False)
```
johandahlberg, line 36: This is super neat.

johandahlberg (approved):
This looks very good. I mostly had some questions, and some very minor comments. Let me know if there i anything you would like to fix, or if I should go ahead and merge

One thought that occurred to me while reviewing this is that I think it would be great to enable test coverage analysis on the PRs, now that we got travis working. What do you think?

withrocks:
It sounds great to enable test coverage.

src/clims/handlers.py
withrocks: Honestly I'm not sure :D I think I might have been referring to the exception type ;)

And you thought I would need weeks to forget it :)

withrocks: ... I'll remove it for now.

src/clims/handlers.py
withrocks: I didn't do it there because I wanted to make sure that the PluginManager was made in the ApplicationService first, to make the lifetime clear. That is, the PluginManger is being used as a global instance which might be created before the ApplicationService.

In the PR I'm working on now, I fix this, by ensuring that the PluginManager is just one of the services created by the ApplicationService and it's a lot cleaner and easier to test.

withrocks: ... so yeah, I broke it out so this could go in sooner.

src/clims/handlers.py
withrocks: Thanks :)

src/clims/handlers.py
withrocks: Yeah sounds good :)

src/clims/handlers.py
withrocks: Right.

src/sentry/testutils/cases.py
withrocks: Good point, I'm fixing this in the current PR so I think it's soon enough that this will actually get fixed.

src/sentry/utils/managers.py
withrocks: Right. In this case I think I might actually have just declared the classes in the InstanceManager itself, and I think it's a bit strange that I didn't. The idea is basically that the exception does not need to be imported, so the user can just do:
```
try:
...
except InstanceManager.ImportException
```

and it's then clear that it's that ImportException the InstanceManager throws (not some other import exception).

Django does this a lot and I think I'm influenced by that. As with so many design patterns, I both like it and I don't. I think it may be clearer to just import the exception as is more usual.

However, I think you might mainly have seen me do this up until now in the plugin-developer-facing code (?) There I think there is an argument for that the user doesn't have to import too much, as it might be a bit easier to get into.

I'm of course fine with changing this if it causes confusion or distress down the line :) But for now I hope it's OK.

withrocks:
Thanks for great feedback! I've fixed some issues and rebased. This can be merged now if you're fine with it.

src/sentry/utils/managers.py
johandahlberg: Thank you for the explanation. I like the argument about the plugin dev not having to import as much. And it's absolutely fine, I'm just educating myself.

johandahlberg:
Cool. This looks good to me. I will merge it when the CI has passed.

PR #120, opened 2020-02-24
Camunda should use postgres again

withrocks:


middleware/camunda/setup.sh
```diff
@@ -1,39 +1,43 @@
...
-# NOTE: Sets up config files and changes the postgres password on every run
+CAMUNDA_VERSION_NO_PATCH=$(python -c "print('.'.join('${CAMUNDA_VERSION}'.split('.')[:2]))")
+
+# NOTE: We install Camunda from source only to get the sql code required to install it. The
+# server is run using the docker image
+CAMUNDA_SERVER_PATH=~/.camunda/server/${CAMUNDA_VERSION}
+mkdir -p ${CAMUNDA_SERVER_PATH}
 
-mkdir -p ~/.camunda
+CAMUNDA_MODELER_VERSION=3.2.3
```
GitEdvard, line 13: I have some thoughts about to split up configuration and code. I have previously in other contexts wrestled with to find where a configuration is set. Perhaps we could find a central place to put all kind of configuration?

scripts/setup-db-user.py
```diff
@@ -51,14 +51,20 @@ def main(role, create_role, print_variables):
     passw = passw.strip()
     if print_variables:
         prefix = role.upper()
-        print("{0}_POSTGRES_DB={1}; export {0}_POSTGRES_DB".format(prefix, role))  # noqa
-        print("{0}_POSTGRES_USER={1}; export {0}_POSTGRES_USER".format(prefix, role))  # noqa
-        print("{0}_POSTGRES_PASSWORD={1}; export {0}_POSTGRES_PASSWORD".format(prefix, passw))  # noqa
+        print("{0}_POSTGRES_DB={1}; export {0}_POSTGRES_DB".format(  # noqa
+            prefix, role))
+        print("{0}_POSTGRES_USER={1}; export {0}_POSTGRES_USER".format(  # noqa
```
GitEdvard, line 56: If I go into stack.yml and ask myself where CLIMS_POSTGRES_USER is set, I will have a hard time finding it. My standard method to find it is to grep it, but that wouldn't work here. Is it necessary with the prefix variable?

GitEdvard (approved):
It look good to me, although I have comments. I think the initialization of clims is generally hard to grip, and has cost me some time to penetrate.

scripts/setup-db-user.py
withrocks: Yeah good point, I know that problem! A case of DRY vs. readability, I guess.

In this case this happens twice, once for `clims` and once for `test_clims`, so it's this or copy the code. It might be good to document it somewhere instead, e.g. in a comment above this. I'll keep that in mind!

withrocks: After thinking about it, I think that in this case readability (or searchability) should weigh more than DRY, but I think I'll just fix it next time I have to work on this, if that's cool.

middleware/camunda/setup.sh
withrocks: It sounds really good to me to move these config values all into the same config file. I'll keep it in mind next time I work on the dev setup!

scripts/setup-db-user.py
withrocks: Thanks for the feedback! I think it would be cool to merge this now and fix this soon (TM).

PR #121, opened 2020-02-25
Stop using the global `plugins` variable

withrocks:
Code should use the PluginManager from the ApplicationService. It's
shared via the global ioc object or by dependency injection.

### DIFF
--- a/src/clims/services/application.py
+++ b/src/clims/services/application.py
@@ -4,7 +4,8 @@ from clims.services.extensible import ExtensibleService
 from clims.services.substance import SubstanceService
 from clims.services.container import ContainerService
 from clims.services.project import ProjectService
-from sentry.plugins import plugins
+from clims.services.workflow import WorkflowService
+from sentry.plugins import PluginManager
 
 
 class ApplicationService(object):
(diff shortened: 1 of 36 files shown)


src/clims/services/application.py
```diff
@@ -4,7 +4,8 @@
 from clims.services.substance import SubstanceService
 from clims.services.container import ContainerService
 from clims.services.project import ProjectService
-from sentry.plugins import plugins
+from clims.services.workflow import WorkflowService
```
GitEdvard, line 7: As far as I understand, since workflow is something configurable, shouldn't it be included by plugins? Perhaps we could have a whiteboard discussion about how the configuration of workflow is thought to happen?

GitEdvard (approved):
This looks good to me!

src/clims/services/application.py
withrocks: Yeah, no doubt, let's do that. This is a part of my major branch that I'm sending in now. The plugins do configure the workflows there.

withrocks:
Missed some essential pieces

PR #122, opened 2020-02-25
Add clims.db for database utility functions

withrocks:


GitEdvard:
I can't see that this code does anything other than raising an exception. Was that the intention? Also, shouldn't it be called somewhere?

withrocks:
Yeah, it's called in a method that I have in another branch. I'm breaking down my crazy-long branch into bitesize pieces.

But yeah, it should just throw an exception if we're already in a transaction, since the functionality that will use it requires data to be written to the DB before continuing.

PR #123, opened 2020-02-25
278 prepare expandable list

GitEdvard:
Purpose:
Prepare to make expandable rows in substance listview. Depending on the value in 'group by' dropdown, decide which api resource to call, either substances or substance property. In substance listview, show mini-arrows in order to expand a grouped row. Also, adjust the rendering of a row depending of if it's grouped or ordinary.

Implementation:
The response from the api call (json) is enhanced with some properties that are added in the reducer. In case of a substance, the field 'isGroupHeader' is added.The substance property comes as a string and is converted to json in the reducer, and the fields 'isGroupHeader', 'id', and 'name' is added. 'Id' is temporarily and is needed to render it in the listview.

Points for careful consideration:
I'm considering to have different action creators/reducer pair for the grouped api call. I'm not satisfied with the current solution since 1) there is a if/else clause on isGroupHeader in both action creator and reducer, which make these two classes tightly coupled. 2) there are rather long blocks of code. I have not settled my mind of this as of yet, because two sets of action/reducers might be heavy, each reducer carries a lot of state. 

Comment:
The large number of rows is due to the update of yarn.lock. I'm not sure why this file was updated in such extent. I think I gave the command 'yarn install'. Should it be included by the version handling?

### DIFF
--- a/src/sentry/static/sentry/app/components/listView.jsx
+++ b/src/sentry/static/sentry/app/components/listView.jsx
@@ -3,6 +3,7 @@ import styled from 'react-emotion';
 import LoadingIndicator from 'app/components/loadingIndicator';
 import LoadingError from 'app/components/loadingError';
 import {Panel, PanelBody} from 'app/components/panels';
+import {CaretDown, CaretRight} from 'app/components/icons';
 import PropTypes from 'prop-types';
 import Checkbox from 'app/components/checkbox';
 
@@ -66,6 +67,33 @@ class ListView extends React.Component {
     listActionBar: PropTypes.object,
   };
 
+  constructor() {
+    super();
+    this.state = {
+      expandedRows: [],
+    };
+  }
+
+  handleRowClick(rowId) {
+    const currentlyExpandedRows = this.state.expandedRows;
+    const isRowExpanded = currentlyExpandedRows.includes(rowId);
+    const newExpandedRows = isRowExpanded
+      ? currentlyExpandedRows.filter(id => id !== rowId)
+      : currentlyExpandedRows.concat(rowId);
+
+    this.setState({expandedRows: newExpandedRows});
+  }
+
+  renderRowExpander(rowId) {
+    const currentlyExpandedRows = this.state.expandedRows;
+    const isRowExpanded = currentlyExpandedRows.includes(rowId);
+    if (isRowExpanded) {
+      return <CaretDown />;
+    } else {
+      return <CaretRight />;
+    }
+  }
+
   getDisplayCell(entryId, header) {
     const row = this.props.dataById[entryId];
 
@@ -94,6 +122,11 @@ class ListView extends React.Component {
             <table>
               <thead>
                 <tr>
+                  {
+                    <th>
+                      <div />
+                    </th>
+                  }
                   {this.props.canSelect && (
                     <th>
                       <Checkbox
@@ -115,6 +148,7 @@ class ListView extends React.Component {
                 {this.props.visibleIds.map(entryId => {
                   return (
                     <tr key={'parent-' + entryId}>
+                      <td>{this.renderRowExpander(entryId)}</td>
                       {this.props.canSelect && (
                         <td>
                           <Checkbox
--- a/src/sentry/static/sentry/app/components/listView.jsx
+++ b/src/sentry/static/sentry/app/components/listView.jsx
@@ -126,11 +126,9 @@ class ListView extends React.Component {
             <table>
               <thead>
                 <tr>
-                  {
-                    <th>
-                      <div />
-                    </th>
-                  }
+                  <th>
+                    <div />
+                  </th>
                   {this.props.canSelect && (
                     <th>
                       <Checkbox
--- a/src/sentry/static/sentry/app/components/listView.jsx
+++ b/src/sentry/static/sentry/app/components/listView.jsx
@@ -6,6 +6,7 @@ import {Panel, PanelBody} from 'app/components/panels';
 import {CaretDown, CaretRight} from 'app/components/icons';
 import PropTypes from 'prop-types';
 import Checkbox from 'app/components/checkbox';
+import {Set} from 'immutable';
 
 const ColumnHeader = styled('div')`
   font-size: 14px;
@@ -84,9 +85,13 @@ class ListView extends React.Component {
     this.setState({expandedRows: newExpandedRows});
   }
 
-  renderRowExpander(rowId) {
+  renderRowExpander(entryId) {
+    const entryData = this.props.dataById[entryId];
+    if (!entryData.isGroupHeader) {
+      return '';
+    }
     const currentlyExpandedRows = this.state.expandedRows;
-    const isRowExpanded = currentlyExpandedRows.includes(rowId);
+    const isRowExpanded = currentlyExpandedRows.includes(entryId);
     if (isRowExpanded) {
       return <CaretDown />;
     } else {
@@ -96,7 +101,6 @@ class ListView extends React.Component {
 
   getDisplayCell(entryId, header) {
     const row = this.props.dataById[entryId];
-
     if (typeof header.accessor === 'function') {
       return header.accessor(row);
     }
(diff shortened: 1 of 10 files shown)


src/sentry/static/sentry/app/components/listView.jsx
```diff
@@ -66,9 +68,39 @@ class ListView extends React.Component {
     listActionBar: PropTypes.object,
   };
 
+  constructor() {
+    super();
+    this.state = {
+      expandedRows: [],
```
withrocks, line 74: I'm not sure if the plan was to keep the expanded rows in here or if it's temporary? My suggestion is to keep state in the components to a minimum, and even keep info on expanded rows in redux, since it makes rationalizing about state much simpler. This approach we use for `visibleIds`, as an example.

src/sentry/static/sentry/app/components/listView.jsx
```diff
@@ -66,9 +68,39 @@ class ListView extends React.Component {
...
+    };
+  }
+
+  handleRowClick(rowId) {
+    const currentlyExpandedRows = this.state.expandedRows;
```
withrocks, line 79: If moved to the reducer, the handling of this state would happen in the reducer.

src/sentry/static/sentry/app/components/listView.jsx
```diff
@@ -94,6 +126,11 @@ class ListView extends React.Component {
             <table>
               <thead>
                 <tr>
+                  {
```
withrocks: I'm not sure if this was added by accident. If this `<th>`  should be here, I think the `{}` should be removed.

src/sentry/static/sentry/app/redux/actions/substanceSearchEntry.js
```diff
@@ -43,21 +48,46 @@ export const substanceSearchEntriesGetFailure = err => ({
...
-      cursor,
-    },
-  };
-  return axios
-    .get('/api/0/organizations/lab/substances/', request)
-    .then(res => {
-      dispatch(substanceSearchEntriesGetSuccess(res.data, res.headers.link));
-    })
-    .catch(err => dispatch(substanceSearchEntriesGetFailure(err)));
+  if (!isGroupHeader) {
```
withrocks, line 59: I believe you were thinking if this should be split into different reducers. I agree with that I think. How about we take a short whiteboard meeting about it?

Regarding that, I don't think you should worry about the amount of data because of multiple reducers. I think that will only be an issue if we start keeping a lot of pages there that the user is not viewing, and then we can just scrap that. Currently we're only keeping one page in the list, and that's tiny amount of data for today's standards.

If we want to focus on making it mobile friendly (with roaming data) that's perhaps different, but that's not a concern in my mind until v2.0+

src/sentry/static/sentry/app/views/substances/substances.jsx
```diff
@@ -82,36 +83,49 @@ class Substances extends React.Component {
       {
         Header: 'Container',
         id: 'container',
-        accessor: d => (d.location ? d.location.container.name : '<No location>'),
```
withrocks, line 85: I hadn't thought much about how headers should be handled in this case, but I have some ideas now. I think it might be nice if we have two header definitions. I think it will be best to take that discussion on a whiteboard meeting.

src/sentry/static/sentry/app/views/substances/substances.jsx
```diff
@@ -220,8 +236,8 @@ const mapStateToProps = state => {
 // TODO: Rename all functions in `mapDispatchToProps` in other files so that they match the action
 // creators name for consistency.
 const mapDispatchToProps = dispatch => ({
-  substanceSearchEntriesGet: (query, groupBy, cursor) =>
-    dispatch(substanceSearchEntriesGet(query, groupBy, cursor)),
+  substanceSearchEntriesGet: (query, groupBy, cursor, isGroupHeader) =>
```
withrocks, line 239: I'd like to just mention - to start the discussion - that I think the action creator names convention should be changed in the future. The actions are named `<noun>_<action>` which seems normal for those constants, but it feels unnatural to me that functions are not named `<action>_<noun>`. In our first redux implementation, this was resolved by mapping back to `<action>_<noun>` in `mapDispatchToProps`, but that adds confusion in my mind.

So tl;dr; I think we should use the more natural function names for all action creators, in this case that would be `getSubstanceSearchEntries`.

tests/js/spec/redux/reducers/substanceSearchEntry.spec.jsx
```diff
@@ -39,25 +39,75 @@ describe('substance reducer', () => {
...
       ...initialState,
       loading: true,
       errorMessage: 'oops',
     };
 
-    const nextState = substanceSearchEntry(prevState, {
+    const action = {
       type: 'SUBSTANCE_SEARCH_ENTRIES_GET_SUCCESS',
-      substanceSearchEntries: mockResponseNoGroup,
+      substanceSearchEntries: JSON.parse(JSON.stringify(mockResponseNoGroup)),
```
withrocks, line 52: Is there a reason for both stringifying and parsing? Isn't it the same result as `mockResponseNoGroup` on its own?

withrocks (commented):
Looks really good. I think we should have a meeting to address those points you brought up.

The `yarn.lock` file should be included and was almost certainly updated because of a `yarn` command, as you say. But I would just remove it from this patch, since it wasn't related to the fix.

tests/js/spec/redux/reducers/substanceSearchEntry.spec.jsx
GitEdvard: The intent of this is to make a deep copy of mockResponseNoGroup. This was actually a nasty thing! The api response is enhanced in the reducer, and I tried to make a failing test for that, but didn't succeed. It turned out that mockResponseNoGroup was mutated (!) via substanceSearchEntries when sent in to the reducer.

src/sentry/static/sentry/app/components/listView.jsx
GitEdvard: This is something that Johan wrote (I overtook his branch), and I haven't reflected on this yet. The functionality for expanded rows is not implemented yet. But, since this will come soon, it's good to reflect on it now! Yes, it could be a good idea to move it to the redux state, although I have not settled my mind of the pros and cons of it yet. The guy that held the redux course at PluralSight claimed that you should not move state to redux unless it was necessary. He wasn't elaborate about it, but that made me curious of why.

src/sentry/static/sentry/app/components/listView.jsx
GitEdvard: Ok, I'll remove that.

tests/js/spec/redux/reducers/substanceSearchEntry.spec.jsx
withrocks: I see. Good that you caught that! That's a bug in the reducer then. Could you add a bug ticket for that?

src/sentry/static/sentry/app/redux/actions/substanceSearchEntry.js
GitEdvard: Yes, I needed some consensus before I split up the reducers. My concern wasn't about performance though, it just felt a little bit bulky to create a new set of redux classes.

src/sentry/static/sentry/app/views/substances/substances.jsx
GitEdvard: Yes, I had the same idea.

src/sentry/static/sentry/app/views/substances/substances.jsx
GitEdvard: Ok!

tests/js/spec/redux/reducers/substanceSearchEntry.spec.jsx
GitEdvard: The state shouldn't be mutable in the reducer! I'll write a ticken on this, and fix it asap.

GitEdvard:
Ok, I think it's ok to merge now. I've taken out the yarn.lock commit (rebased), and fixed a minor bracket thing. I'll write a ticket on having immutable state in the reducer and fix that next.

PR #124, opened 2020-02-28
Add a new logo and related UI changes

withrocks:


src/sentry/static/sentry/app/views/app.jsx
```diff
@@ -222,7 +222,7 @@ const App = createReactClass({
   render() {
     if (this.state.loading) {
       return (
-        <LoadingIndicator triangle={true}>
+        <LoadingIndicator>
```
GitEdvard, line 225: I think the swirling triangle was appealing. But maybe we shouldn't use the obvious sentry stuff?

LICENSE
```diff
@@ -2,6 +2,17 @@ Alterations of Sentry for Common LIMS:
...
+
+    2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+
+    3. Neither the name Common LIMS, nor the Common LIMS logo nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
```
GitEdvard, line 15: I'm curious of where you got this formulation from. Is this a standard license text?

GitEdvard (approved):
This looks ok for me!

src/sentry/static/sentry/app/views/app.jsx
withrocks: I agree that it looks cool, but it's their logo.

We could easily make a swirling icon of our own later though!

LICENSE
withrocks: Right! It's the BSD 3-clause license. We've had it from the beginning as it's what Sentry originally had when we forked it.

What I've done here is broken it down into their original license and our license. Our license replaces "Sentry" with "Common LIMS" and also adds the sentence part "nor the Common LIMS logo" in the third part.

The reason for that addition is that I want to make it explicit that we're not relicensing our logo, as I'm not sure we can (and there's no reason to either). I (costeinar@gmail.com) bought a license of it for use in this project from the Noun project and the requirements state:

"Once you purchase the icon you may use it as many times as you would like as long as it is for the same project/client and you may use it in any medium."

So by adding this clause, I believe that we've made sure that that the logo can be used by anyone using Common LIMS, but not if they create a derived product.

GitEdvard: Ok!

PR #125, opened 2020-03-04
Downgrade python to 2.7.16 for travis

withrocks:
This is a temporary downgrade as builds are currently failing


withrocks:
Didn't work, evaluating github actions instead.

PR #127, opened 2020-03-05
clims-0, fix create example data

GitEdvard:
Purpose:
Place samples in containers (this seems to be lost during some refactoring?). Also, fix so that createexampledata can be run on its own, without being called within make fresh. There was a problem with plugin registration of the Demo plugin, and registration of Plate96.

Comments:
Maybe we could find something more general than Plate96?

### DIFF
--- a/src/clims/plugins/demo/handlers/create_example_data.py
+++ b/src/clims/plugins/demo/handlers/create_example_data.py
@@ -4,7 +4,9 @@ import logging
 from uuid import uuid4
 
 from clims.handlers import CreateExampleDataHandler
-from clims.plugins.demo.models import ExampleSample, ExampleProject
+from clims.plugins.demo.models import ExampleSample, ExampleProject, Plate96
+from clims.plugins.demo import DemoPlugin
+from clims.models.plugin_registration import PluginRegistration
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


src/clims/plugins/demo/handlers/create_example_data.py
```diff
@@ -18,6 +20,14 @@ def handle(self):
         Any plugin that implements this handler will supply demo data
         """
 
+        # In case create example data is run on its own, (not within make fresh)
+        # the plugin registration is needed
+        demo_plugin_full_name = '{}.{}'.format(DemoPlugin.__module__, DemoPlugin.__name__)
```
withrocks, line 25: I think the reason it works when you're running `make fresh` is that `make fresh` also runs `lims upgrade`. The idea is that when devs want new plugin code they must do two things: (1) pip install the plugin package (2) run `lims upgrade`. This registers all new types.

Would you mind removing this code and see if it works to just run `lims upgrade` before creating demo data?

withrocks:
This looks good! But I'm pretty sure we don't need the registration in the example code. It's a bug if running `lims upgrade` doesn't take care of it.

This might be unclear in the documentation still, any comments on that are very welcome :)

src/clims/plugins/demo/handlers/create_example_data.py
withrocks: This is a test comment. Checking if our slack integration picks it up!

PR #132, opened 2020-03-10
Refactor plugin manager

withrocks:
This mainly focuses on stop using the global `plugins` singleton. Also includes some fixes for the plugin manager.

Simplifies somewhat the loading of plugins.

This is partially the same as the following closed PR that was already code reviewed by @GitEdvard. So reviewing the commit 925f241 should suffice.

https://github.com/commonlims/commonlims/pull/121

I closed that PR as it was missing a few pieces that were related to the second commit.

### DIFF
--- a/src/sentry/plugins/base/v2.py
+++ b/src/sentry/plugins/base/v2.py
@@ -12,7 +12,9 @@ __all__ = ('Plugin2', )
 import logging
 import six
 import inspect
-import os
+import sys
+import importlib
+import pkgutil
 
 from django.http import HttpResponseRedirect
 from threading import local
@@ -27,6 +29,7 @@ from sentry.plugins.base.configuration import (
     default_plugin_options,
 )
 from sentry.utils.hashlib import md5_text
+from clims.services.extensible import ExtensibleBase
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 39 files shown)


src/sentry/plugins/base/v2.py
```diff
@@ -480,17 +483,69 @@ def handle_signal(self, name, payload, **kwargs):
...
+
+    @classmethod
+    def _find_subclasses_of(cls, parent_cls, module):
+        """
+        Loads and finds all implementations of a particular subclass in a class module or any
+        of its submodules. Used to find workflows, handlers etc. defined in the plugin.
+        """
+
+        def is_disallowed_core_class(module_name):
+            # The only "core" classes that should ever be registered are those in the
```
GitEdvard, line 510: I think it should be "never" instead of "ever"

src/sentry/plugins/base/v2.py
```diff
@@ -480,17 +483,69 @@ def handle_signal(self, name, payload, **kwargs):
...
+            logger.debug("Subclasses of '{}' in '{}': {}".format(parent_cls, mod_name, clsmembers))
+            return clsmembers
+
+        ret = set()
+        # Search in the current module and all submodules
+        ret.update(find_implementations_in(module.__name__, parent_cls))
+        if hasattr(module, "__path__"):
+            # This is a package
+            for _, mod_name, _ in pkgutil.walk_packages(path=module.__path__, prefix=module.__name__ + "."):
+                ret.update(find_implementations_in(mod_name, parent_cls))
```
GitEdvard, line 537: Cool, it seem like the extensible types is no longer restricted to the "models.py" module.

src/sentry/plugins/base/v2.py
```diff
@@ -480,17 +483,69 @@ def handle_signal(self, name, payload, **kwargs):
...
+        return ret
+
+    @classmethod
+    def get_extensible_objects(cls):
```
GitEdvard, line 541: Perhaps the name should be "get_extensible_classes" or something? Objects make me think of instances of a class.

src/sentry/plugins/base/v2.py
```diff
@@ -480,17 +483,69 @@ def handle_signal(self, name, payload, **kwargs):
...
-                    yield os.path.join(workflows_dir_path, file_path)
-        except OSError:
-            pass
+            import importlib
+            return importlib.import_module(module_name)
+        except ImportError as ex:
+            if six.text_type(ex) != "No module named {}".format(name):
+                trace = sys.exc_info()[2]
+                raise ImportError("Error while trying to load plugin {}".format(module_name)), None, trace
+            logger.debug("Can't find module {}".format(module_name))
```
GitEdvard, line 500: I think it might be more expressive to also write "return None" here.

GitEdvard (approved):
This looks good!

src/sentry/plugins/base/v2.py
withrocks: Yeah I agree! I think I'll fix it in the next iteration :+1:

withrocks: I can throw it into my next PR.

src/sentry/plugins/base/v2.py
withrocks: Yeah, seems I'm all over the place in the nomenclature here

src/sentry/plugins/base/v2.py
withrocks: Right!

src/sentry/plugins/base/v2.py
withrocks: I was actually thinking "ever", sounds a bit dramatic now that I read it :)

PR #133, opened 2020-03-11
Code generator

withrocks:
We've got a lot of code that contains a lot of boilerplate. This could be addressed dynamically on runtime (e.g. generate functions on runtime based on configuration), but a straightforward way is to generate code, which in turn has some benefits when it comes to readability and debugging.

This is the first draft (functioning) of a tool that takes the latter approach, which for now generates just a hardcoded "resource" (think of a REST resource).

My idea is that the tool should be run like this:

`lims django codegen [resource]`

where `resource` is e.g. `Task` or `User`

This will generate all the redux code for the resource, but it could also create placeholder views etc to guide the developer in what's required for new resources.

### DIFF
--- a/src/clims/devhelpers/generate.py
+++ b/src/clims/devhelpers/generate.py
@@ -1,0 +1,93 @@
+from __future__ import absolute_import
+
+import os
+
+from clims.devhelpers import resources
+import re
+from jinja2 import Template
+import stringcase
+
+# TODO: This should be a separate module (e.g. clims_devtools)
+
+
+class CodeGenerator(object):
+    """
+    Generates starting point code for the CLIMS project. Given a resource name, generates:
+        * Redux actions and reducer
+        * [TODO not implemented yet] API endpoint for listing the resource
+        * [TODO not implemented yet] API endpoint for details
+    """
+
+    def __init__(self):
+        self.actions_path = "src/sentry/static/sentry/app/redux/actions"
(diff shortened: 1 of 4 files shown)


src/clims/devhelpers/generate.py
```diff
@@ -0,0 +1,93 @@
...
+import stringcase
+
+# TODO: This should be a separate module (e.g. clims_devtools)
+
+
+class CodeGenerator(object):
+    """
+    Generates starting point code for the CLIMS project. Given a resource name, generates:
+        * API endpoint for listing the resource
+        * API endpoint for details
```
johandahlberg: I guess these are not true yet? All I see so far is code to generate the redux actions.

johandahlberg (commented):
I like this. Seems like to could be really useful. Perhaps clean up the docs to make some things TODO which are not yet done? Also I see that there is an empty file `get.j2`? Has that been accidentally committed?

src/clims/devhelpers/generate.py
withrocks: Yeah, these are just empty promises at the moment :P I'll throw in TODOS.

withrocks:
I think you're right about that `get.j2` file. Will fix.

withrocks:
Fixed.

johandahlberg:
Waiting for CI to be green, then I'll merge.

PR #134, opened 2020-03-11
Introduce the ResultIterator

withrocks:
The `ResultIterator` behaves much like the part of a Django `QuerySet` that
deals with iteration. We use it to iterate over results of thick models,
which are based on Django models.

### DIFF
--- a/src/clims/services/base_extensible_service.py
+++ b/src/clims/services/base_extensible_service.py
@@ -6,6 +6,7 @@ from django.core import exceptions as django_exceptions
 from clims.services.exceptions import DoesNotExist
 from clims.models.extensible import ExtensibleProperty
 from clims.services.extensible import ExtensibleTypeNotRegistered
+from clims.models import ResultIterator
 
 
 class BaseExtensibleService(object):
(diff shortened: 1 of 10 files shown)


src/clims/services/base_extensible_service.py
```diff
@@ -61,24 +62,24 @@ def all(self):
         # which in turn would return the wrapper when materialized. For that to work smoothly,
         # we'll need to look into implementation details of django querysets.
 
-        # TODO: how does the prefetch perform when fetching all objects like this
-        for entry in self._all_qs():
-            yield self.to_wrapper(entry)
+        # TODO: The `to_wrapper` method is NOT on this mixin, which makes this difficult to reason
```
johandahlberg, line 65: I agree with this. Do you think we should fix it in this PR, or do it later?

johandahlberg (approved):
This looks very nice and clean to me. Good job! Do you want to fix the TODO I commented on now? Or wait and do it later?

src/clims/services/base_extensible_service.py
withrocks: I think I would like to leave it for later, as I has a ton of PRs to get in :|

withrocks:
Cool, thanks! Please merge it.

PR #135, opened 2020-03-11
Introduce the thread-local context

withrocks:
One of the main goals of Common LIMS is to provide a very simple way to
extend the system. The most basic way to extend the system is in plugin
handlers. The plugin dev should have the power to do everything, but it
should be simple, at least in most cases.

On of the pieces of that design is the thread-local context. Until now
we've just had the "context", which is the same thing but I waited with
implementing this part of it, which is how objects share the context.

The context has:
- the application service, from which all other services can be derived
- the organization
- the current user
(more things may come)

When a handler executes, a context has already been started for the
user. It effectively enhances the business object, because they now have
access to everything the plugin dev requires, without the dev having to
initialize the objects with that data. This means that the user can do
this:

    sample = OurSample(name="testname")
    sample.save()

and it will work, even if they didn't add any other required info, such
as the user or the organization.

A very similar approach *would* have been allowing the user only to
create objects through a factory, e.g.:

    sample = self.OurSample(name="testname")

but since one of the core goals of the plugin system is readability, I
chose the previous approach.

This approach is similar to Django's atomic db transactions, which can be started with
e.g. `with transaction.atomic()`. Django transactions do also use thread local storage to
communicate data between different parts of the code.

### DIFF
--- a/src/clims/services/extensible.py
+++ b/src/clims/services/extensible.py
@@ -7,6 +7,7 @@ from clims.models.extensible import (ExtensibleType, ExtensiblePropertyType, Ext
 from clims.models.location import SubstanceLocation
 from django.db import transaction
 from clims import utils
+from clims.handlers import context_store
 
 
 logger = logging.getLogger(__name__)
@@ -26,13 +27,13 @@ class ExtensibleService(object):
     def clear_cache(self):
         self._model_cache.clear()
 
-    def create(self, name, klass, organization, properties=None):
+    def create(self, name, klass, properties=None):
         """
         Creates an extensible and saves it to the backend.
         """
         if isinstance(klass, six.string_types):
             klass = self.get_implementation(klass)
-        instance = klass(name=name, organization=organization, _app=self._app)
+        instance = klass(name=name)
 
         if properties:
             for key, val in properties.items():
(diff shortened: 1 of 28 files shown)


src/clims/services/extensible.py
```diff
@@ -220,25 +222,27 @@ def __init__(self, **kwargs):
...
         extensible_type = self._app.extensibles.get_extensible_type(self.type_full_name)
         self._archetype = self.WrappedArchetype(name=name, extensible_type=extensible_type,
-                organization=org)
+                organization=self._context.organization)
         self._wrapped_version = self.WrappedVersion()
 
         # Add any remaining properties in kwargs. This is necessary so that user
         # can instantiate objects using e.g. syntax like: Sample(my_value=1)
         for key, value in six.iteritems(kwargs):
+            assert key != "organization"
```
johandahlberg, line 236: Is this a debug thing that hasn't been removed?

src/sentry/testutils/fixtures.py
```diff
@@ -787,3 +788,14 @@ def create_container_with_samples(
             container.append(sample)
         container.save()
         return container
+
+
+def has_context(func):
```
johandahlberg, line 797: This does not appear to be used?

johandahlberg (commented):
This looks good, and my comments are minor. However, I am not sure I follow exactly how this is used. Most likely I am missing something.

What I don't understand specifically is when you use the `CreateContext` content handler. I would expect it to be used e.g. when creating handlers, but now I can only see it being used in the test code and the code to create example data. Am I misunderstanding something here, or is something missing?

src/sentry/testutils/fixtures.py
withrocks: Right! I added it so one can add a context around a test function without the test having a context. I then went with having it defined in the core test case as it's more convenient.

src/clims/services/extensible.py
withrocks: Yes! I was going to remove it after I found all the organizations.

withrocks:
Thanks for the review. You're point about this not being called before the handlers is right! I hadn't started on the tests for plugins so I missed adding it. I'll throw it in now so we can start using this pattern also in the plugins.

The failed tests have to do, I believe, with the order of when fixtures are evaluated in the tests. I'll fix that before we can merge!

withrocks:
WIP lifted!

PR #136, opened 2020-03-11
clims-351, fix create example data

GitEdvard:
Purpose:
1) In create-example-data, put back the creation of containers (seem to be lost during a refactoring), and place the samples in them. 
2) Fix a bug that hindered the rendition of samples in UI. This bug kicked in when samples were fallbacked from plugin classes to base classes, e.g. instantiated as SubstanceBase instead of DemoSample. 
3) Fix a bug that made samples and container instantiated as base classes in ui (SubstanceBase and ContainerBase). 

Implementation:
All extensible types is loaded in working memory (in ExtensibleService). This loading is now done lazily and fetched from database, so that every new instance of ExtensibleService is garanteed to be properly initialized, which was not the case before this bugfix.

Note:
I have not taken into account which versions of the extensible types that is loaded. I couldn't find the field "latest", so I postphoned this until later.  

This PR is a modified version of an old PR, which I closed: https://github.com/commonlims/commonlims/pull/127

### DIFF
--- a/src/clims/services/extensible.py
+++ b/src/clims/services/extensible.py
@@ -21,7 +21,7 @@ class ExtensibleService(object):
     def __init__(self, app):
         self._app = app
         self._model_cache = dict()
-        self._implementations = dict()
+        self._implementations = None
         self._logger = logging.getLogger(self.__class__.__name__)
 
     def clear_cache(self):
@@ -97,6 +97,8 @@ class ExtensibleService(object):
         """
         This is for testing purposes only!
         """
+        # trigger initiation of the implementation array
+        self.implementations
         name = '{}.{}'.format(extensible_base.__module__, extensible_base.__name__)
         del(self._implementations[name])
 
--- a/src/clims/services/extensible.py
+++ b/src/clims/services/extensible.py
@@ -87,6 +87,13 @@ class ExtensibleService(object):
                                                property_types=property_types)
         return extensible_type
 
+    def unregister_model(self, extensible_base, plugin):
+        """
+        This is for testing purposes only!
+        """
+        name = '{}.{}'.format(extensible_base.__module__, extensible_base.__name__)
+        del(self._implementations[name])
+
     @transaction.atomic
     def _register_model(self, name, org, plugin, property_types=None):
         """
--- a/src/clims/services/extensible.py
+++ b/src/clims/services/extensible.py
@@ -20,8 +20,8 @@ class ExtensibleTypeNotRegistered(Exception):
 class ExtensibleService(object):
     def __init__(self, app):
         self._app = app
-        self._implementations = dict()
         self._model_cache = dict()
+        self._implementations = dict()
         self._logger = logging.getLogger(self.__class__.__name__)
 
     def clear_cache(self):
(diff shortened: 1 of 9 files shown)


withrocks:
The `latest` flag I mentioned is on the `PluginRegistration` model, but in a PR that I'll open later today :|

src/clims/plugins/demo/handlers/create_example_data.py
```diff
@@ -30,9 +28,25 @@ def handle(self):
...
 
-        available_sample_types = ["Vampire Fang", "Zombie Brain", "Hydra Claw"]
+        available_containers = []
         for ix in range(100):
-            name = "demoplugin-sample-{}".format(ix + 1)
+            id = ix + 1
+            # id = uuid4().hex
+            name = 'pandora-{}'.format(id)
+            try:
+                container = self.app.containers.get(name=name)
```
withrocks: I think we should add a `get_or_create` method to `BaseExtensibleService` to support this in one method call, copying Django's semantics for the method with the same name. Excellent if you could do that here, but it can also wait.

src/clims/services/extensible.py
```diff
@@ -77,11 +87,19 @@ def register(self, plugin_reg, extensible_base):
...
 
         name = "{}.{}".format(extensible_base.__module__, extensible_base.__name__)
-        self._implementations[name] = extensible_base
         extensible_type = self._register_model(
             name, 'default', plugin_reg, property_types=property_types)
         return extensible_type
 
+    def unregister_model(self, extensible_base, plugin):
+        """
+        This is for testing purposes only!
```
withrocks: I would prefer that code that's just for tests is kept in the tests only. I believe that keeping only what's required in the API makes it simpler to manage, e.g. in terms of knowing what requires tests.

src/sentry/plugins/base/manager.py
```diff
@@ -117,7 +116,21 @@ def install_extensible_types(self, plugin):
...
+    def get_extensible_types_from_db(self):
+        """
+        :return: class objects of extensible types
+        """
+        from clims.models.extensible import ExtensibleType
+
+        extensible_types = ExtensibleType.objects.all()
+        type_names_in_db = [e.name for e in extensible_types]
+        for type_name in type_names_in_db:
+            splitted = type_name.split('.')
```
withrocks: This should just be `split` not `splitted`.

src/sentry/plugins/base/manager.py
```diff
@@ -337,15 +350,20 @@ def get_plugin_module(self, plugin_class, name):
         """
         Gets a module defined in the plugin. Returns None if it wasn't found
         """
-        module_name = "{}.{}".format(plugin_class.__module__, name)
+        full_module_name = '{}.{}'.format(plugin_class.__module__, name)
+        return self._import_module(full_module_name)
 
+    def _import_module(self, full_module_name):
+        splitted = full_module_name.split('.')
```
withrocks: ... same here, s/splitted/split.

withrocks (commented):
This looks good to me. I have a few comments.

withrocks:
There are some conflicts now.

src/clims/plugins/demo/handlers/create_example_data.py
GitEdvard: I opt for waiting with this. In order to do this, container service has to be initialized with the specific class that needs to be instantiated (i.e. HairSampleContainer). But I'm unsure how the behaviour of container service is thought of to be. What if for example there is more than one container implementation floating around in one session?

src/sentry/plugins/base/manager.py
GitEdvard: Ok

PR #137, opened 2020-03-12
Container should support iteration

withrocks:
Now one can iterate through the container with:

    `for ix, item in container`

This returns both the index object as well as the item.

One can also get the contents of the container (non-empty) with:

    container.contents


johandahlberg (approved):
Nice!

PR #138, opened 2020-03-19
Workflows

withrocks:
- Implements plugin defined workflows
- Some related and unrelated refactoring


setup.cfg
```diff
@@ -15,6 +15,9 @@ exclude = .venv/.git,src/*/migrations/*,node_modules/*,src/sentry/static/sentry/
 [bdist_wheel]
 python-tag = py27
 
+[yapf]
```
johandahlberg, line 18: I don't remember - is there any pre-commit auto-formatting done?

src/clims/api/endpoints/process_assignments.py
```diff
@@ -0,0 +1,26 @@
...
+class ProcessAssignmentsEndpoint(Endpoint):
+    authentication_classes = (SessionAuthentication, )
+    permission_classes = (IsAuthenticated, )
+
+    def post(self, request, organization_slug):
+        """
+        Assign one or more item to a workflow. The items are assigned by global_id.
+        """
+
+        # TODO-auth: Ensure that the user is only assigning samples that are under the organization
```
johandahlberg, line 18: I think it would be good to gather all these auth questions up into an epic and go over the systematically, so that they are not forgotten.

src/clims/models/assignment.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import absolute_import
+
+from django.db import models
+from sentry.db.models import (FlexibleForeignKey, sane_repr)
+from jsonfield import JSONField
+
+
+class SubstanceAssignment(models.Model):
```
johandahlberg, line 8: If this is only relevant for user actions, perhaps it would be good to indicate that in the name. I'm thinking that perhaps this could be called something like `UserSubstanceAssignment`? What do you think?

src/clims/models/assignment.py
```diff
@@ -0,0 +1,44 @@
...
+    __core__ = True
+
+    substance = FlexibleForeignKey('clims.Substance', related_name="assignments")
+
+    user = FlexibleForeignKey('sentry.User')
+
+    workflow = FlexibleForeignKey('clims.Workflow', null=True)
+
+    # The preset of variables used when assigning
+    preset = models.TextField()
```
johandahlberg, line 34: How are these used? I'm not sure I understand the distinction between this and `variables`.

src/clims/api/serializers/models/task.py
```diff
@@ -0,0 +1,19 @@
...
+
+from clims.api.serializers.models.substance import SubstanceSerializer
+from clims.services.substance import SubstanceBase
+
+
+class TaskSerializer(serializers.Serializer):
+    id = serializers.CharField()
+    tracked_object = serializers.SerializerMethodField()
+
+    def get_tracked_object(self, obj):
```
johandahlberg: How is this method used? I can't see it being used elsewhere in the code (though I might have simply missed it with just searching in the Github interface), nor as being part of the Serializer APIs.

src/sentry/testutils/cases.py
```diff
@@ -404,6 +404,11 @@ def assertWriteQueries(self, queries, debug=False, *args, **kwargs):
         with context:
             func(*args, **kwargs)
 
+    def clean_workflow_engine_state(self):
+        from clims.services.workflow.camunda import CamundaClient
+        client = CamundaClient(self.app.settings.CAMUNDA_API_URL)
+        client.unsafe_delete_all_deployments()
```
johandahlberg, line 410: Lets hope no one gets the idea to hock up the tests to their prod db. :wink:

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+
+    @property
+    def definition_id(self):
+        return self.get_full_name()
+
+    def get_available_for_workbatch(self):
+        """
+        Returns all items that are available for workbatch of this type.
+        """
+        # print(self.get_full_name(), self.available_work)
```
johandahlberg: Looks like there might be a missing TODO here?

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+    def preset(self, name, **kwargs):
+        """
+        Creates a preset with the name and kwargs and then adds it to the workflows presets
+        """
+        new_preset = create_preset(name, **kwargs)
+        self._presets.update(new_preset)
+
+    def get_presets(self):
+        """
+        Fetches the presets defined in a subclass either with a dictionary called `presets`
```
johandahlberg, line 112: Why might the user (I guess in this case a plugin dev) want to define their own private `_presets` instead of using the regular mechanism? I think it would be useful to add that to the docs to guide folks through this.

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+
+        from .camunda import CamundaWorkflow
+        from .camunda.client import CamundaClient
+
+        # A dict of handlers that proxy an external workflow management system:
+        self.handlers = {
+            "camunda": {
+                "key": "camunda",
+                "cls": CamundaWorkflow,
+                "handler": CamundaClient(self._app.settings.CAMUNDA_API_URL)
```
johandahlberg, line 213: Is there any need to pass on credentials etc here as well? I guess the current Camunda instance is not locked down, but I guess it would be in a production env?

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+
+        # TODO(clims-345): Make sure that we're not adding a substance to another workbatch if
+        # it's already "locked".
+        return batch
+
+    def get_tasks_by_ids(self, task_ids):
+        task_ids_by_provider = defaultdict(list)
+
+        for task_id in task_ids:
+            provider_id, id = task_id.split("/")
```
johandahlberg, line 296: This might relate more to me not understanding the API of the `create_work_batch` method. Why is there a need for a split here? I guess they ids have a format where they define a provider and then their their actual id? What does the provider mean in this context?

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+    def assign_item(self, workflow, item, user):
+        from clims.models import SubstanceAssignment
+
+        # Fetch the handler that takes care of creating the workflow in the external workflow engine
+        handler = self.get_handler(workflow.__class__)
+        assignment = self.AssignmentModel(item, user)
+
+        # Fetch the model we use to keep a history of the item
+        # logger.info("Assigning {} to {} with vars {}".format(
+        #     item, workflow_class, {}))
```
johandahlberg: Should this be removed?

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+
+        # NOTE: We don't want the assignment to be saved in a transaction. The reason is that
+        # we want an uncompleted item in the assignment table to be able to identify workflows
+        # we might have started or not, of which we don't have any record. This is necessary
+        # since the workflow engine is outside of our database transactions.
+        require_no_transaction()
```
johandahlberg: How does this work? And for how long is it active? Since it is not a context handler it is not obvious from the code when transactions are required again.

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+
+    def get_task_definitions(self, process_definition_key, task_definition_key):
+        from clims.models import CamundaTask
+        from django.db.models import Count
+        from django.db.models import Q
+
+        # TODO: Only supports camunda
+        task_definitions = CamundaTask.objects.select_related('process_definition')
+
+        # Temporary: Remove Camunda demo data results:
```
johandahlberg, line 509: Do we want a TODO here as well, or maybe a ticket mentioning to clean it out when we have a proper process to populate the test camunda instance up and running?

src/clims/services/workflow/camunda/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+        """
+        Returns the .bpmn file that further defines the workflow or subworkflow.
+        This file is what's uploaded to the external workflow engine.
+        """
+        module_path = os.path.dirname(inspect.getfile(inspect.getmodule(cls)))
+        logger.debug("Searching for .bpmn file at {}".format(module_path))
+
+        for file_path in os.listdir(module_path):
+            fname, ext = os.path.splitext(file_path)
+            if ext == ".bpmn" and fname == cls.__name__:
```
johandahlberg, line 30: Probably no big thing, but perhaps we also want to do a check that it actually is bpmn file as well? I'm guessing that the xml files contain some type of header declaring their type?

src/clims/services/workflow/camunda/api.py
```diff
@@ -117,15 +135,30 @@ class ProcessDefinition(CamundaResource):
...
         self.url = url
 
-    process_definition = ResourceOfType(ProcessDefinition, "process-definition/{id}")
+    process_definition = ResourceOfType(ProcessDefinition,
+                                        "process-definition/{id}")
 
     process_instance = ResourceOfType(ProcessInstance, "process-instance/{id}")
     process_instances = ResourcesOfType(ProcessInstance, "process-instance")
+
+    deployment = ResourceOfType(Deployment, "deployment/{id}")
```
johandahlberg, line 160: I realize that a deployment is some kind of camunda concept. Could you explain what it maps to? I took a quick look at the camunda docs, but I didn't understand it.

src/clims/services/workflow/camunda/client.py
```diff
@@ -0,0 +1,370 @@
...
+from clims.models import Workflow
+
+logger = logging.getLogger(__name__)
+
+
+class CamundaClient(object):
+    """
+    A client for executing workflows that interfaces with the underlying workflow engine.
+    It's higher level than CamundaApi which directly maps to the CamundaApi.
+    """
```
johandahlberg, line 21: In general I'm wondering if we want to put some work into this class to ensure that it retries automatically, etc. I saw that you had brought in the retry library, so I suppose that should be fairly easy to get going here?

src/clims/models/workflow.py
```diff
@@ -0,0 +1,43 @@
...
+    # True if this is the latest version of this workflow definition (with this definition_key)
+    latest = models.BooleanField()
+
+    # The backend type used
+    backend = models.TextField()
+
+    # The plugin (down to the version) that registered the workflow
+    plugin_registration = FlexibleForeignKey('clims.PluginRegistration')
+
+    # organization = FlexibleForeignKey('sentry.Organization', related_name='workflows')
```
johandahlberg, line 37: Should workflows not be tied to organizations?

src/clims/api/endpoints/work_batch_details.py
```diff
@@ -0,0 +1,13 @@
+from __future__ import absolute_import
+
+from rest_framework.response import Response
+from clims.api.bases.work_batch import WorkBatchBaseEndpoint
+from clims.models import WorkBatch
+from clims.api.serializers.models.workbatch import WorkBatchSerializer
+# from sentry.api.paginator import OffsetPaginator
```
johandahlberg: Remove commented import?

tests/clims/api/endpoints/test_work_batch.py
```diff
@@ -0,0 +1,73 @@
...
+            "tasks": tasks,
+        }
+
+        response = self.client.post(
+            path=url,
+            data=json.dumps(payload),
+            content_type='application/json',
+        )
+        assert response.status_code == 201
+        print(response.json())
```
johandahlberg: Forgotten print statement?

src/clims/plugins/demo/dnaseq/workflows/sequence.py
```diff
@@ -0,0 +1,77 @@
+from __future__ import absolute_import
+
+# TODO: Should we have a namespace from which plugin developers should import
```
johandahlberg, line 3: Sounds like a good idea to make it easier for developers to know that they are getting "the right stuff".

src/clims/plugins/demo/dnaseq/workflows/sequence.py
```diff
@@ -0,0 +1,77 @@
...
+# (e.g. only clims.plugins)?
+
+from clims.services.workflow.camunda import CamundaWorkflow
+from clims.services import TextField
+
+SAMPLE_TYPES = ['DNA', 'RNA']
+PREP_METHODS = ['microwave', 'mixer']
```
johandahlberg, line 10: I always prep my DNA in the microwave! :smile:

src/clims/plugins/demo/dnaseq/workflows/sequence.py
```diff
@@ -0,0 +1,77 @@
...
+
+class SequenceSimple(CamundaWorkflow):
+    """
+    Sequence a sample.
+
+    Note that since this is a `CamundaWorkflow`, the flow of the workflow is described in
+    a file with the same name as the class but the .bpmn extension. This class acts as glue between
+    the visual workflow definition and the data entry views that are required for completing the
+    workflow.
+    """
```
johandahlberg, line 22: Its very nice to see this. It really shows how it is all coming together beautifully in the user-facing APIs. Splendid!

src/clims/services/extensible.py
```diff
@@ -191,40 +233,112 @@ def __set__(self, obj, value):
         obj._property_bag[self.prop_name] = value
 
 
-class PropertyInitiator(type):
+class HasExtensibleFieldsMeta(type):
+    """
+    A meta class for classes that have extensible fields. Classes that have this metaclass
+    can desribe extensible statically on the object with this syntax:
```
johandahlberg: desribe -> describe

johandahlberg (approved):
I've gone over this now. Though give how much it is I cannot say I have done more than a shallow review at this stage. In particular I have only glanced at the front-end changes, since I have a particularly difficult time following those given my limited experience.

It all looks really good to me, and I'm happy to see this coming together! My comments are mostly to clarify this to myself, or suggestions for tickets to create. If you want me to I'm good to merge this.

Great work!

setup.cfg
withrocks: Yes there is. It's actually using flake, I believe, with autopep8 settings (it's code I haven't had to look at much), but it's limited to for example not formatting dictionaries so I added this setting so I can format files with yapf.

There is a third player in this, black, which I think might be the best tool for this, but it requires py3 so it has to wait a while. I figured that we might revisit this a bit when we can use that.

src/clims/api/endpoints/process_assignments.py
withrocks: Yes, I agree. We already have an epic for it, where the first task is to find all these todos. I've clarified that in the ticket and moved it up in the epic list so it's more visible

src/clims/models/assignment.py
withrocks: I believe you might be thinking about my comment to the class that this only has to do with human interactions? I think that it might be indicated by that these are "assignments" to a workflow that they must be made by a user, but I wanted to be clear on that this model doesn't have information on what happened in the workflow after that happened. I might have to revisit that documentation to clarify it.

So to clarify, there is no assignment made by an entity that's not a user, so I guess it might be unnecessary to say that it's a user because it's implied.

src/clims/models/assignment.py
withrocks: Presets can be thought of as a named tuple of `(process, variables)`. They are only there for the end user, so they don't have to specify variables that almost always might be the same.

It's similar to having a function for end users of a library that calls a lower level function that does all the work, but is too inticrate for end users, e.g.:

```
def _print(self, a, b, c, color, e, f, g, **kwargs): ...
 
def print_color(self): ...

def print_bw(self): ...
```

This variable in the model is just to know which preset was selected (the name of it) so it can be shown in the UI.

src/clims/api/serializers/models/task.py
withrocks: I believe you're right!

tests/clims/api/endpoints/test_work_batch.py
withrocks: Oops. I think our linter must then not complain about it if its in the tests.

withrocks:
Thanks for the review! I fixed the issues and rebased (because the fixes were so small). This can be merged now in my mind.

src/clims/models/workflow.py
withrocks: I feel like we might need to revisit that question later. I hit some unanswered questions when it comes to relating plugins and organizations. When you install a plugin today, they are available to all organizations, because I wasn't sure how we would want to solve it from a user perspective - it's not as simple as pip install in that case. Since the workflows follow directly from the plugin, they don't have a connection to the workflow now, because plugins are on org level.

src/clims/services/workflow/camunda/__init__.py
withrocks: Yeah, that might be good to have. There will be validation of it though when we try to register it in Camunda, so for simplicity we coul perhaps live with that?

src/clims/services/workflow/camunda/api.py
withrocks: Right, deployment is when you "deploy" one or more workflow definitions to the system. It's like a commit.

src/clims/services/workflow/camunda/client.py
withrocks: Right. The retry library was only brought in for tests. I think if we want to enable retries (which I think is a very good thing of course) for communication with external resources, we might perhaps want to do that with a message server?

I'll add a ticket for this, great point!

withrocks: But I'd also like to add that the Camunda server is going to be under little load in most setups (unless someone wants to use this to host a lims system for several organizations at once, which doesn't need to be in scope for now I guess). And the server is also running locally, so I'd argue that retries are not as important here as they would be if this was "truly" an external system. But I still think we should discuss it!

src/sentry/testutils/cases.py
withrocks: Yeah :) I guess they might expect strange things to happen in that case.

src/clims/models/assignment.py
johandahlberg: :+1:

src/clims/models/assignment.py
johandahlberg: Yes, I think this became clearer to me as well, as I read more of the code.

src/clims/services/workflow/camunda/__init__.py
johandahlberg: Yeah, that sounds reasonable. :+1:

src/clims/services/workflow/camunda/api.py
johandahlberg: Aha. I get it.

src/clims/services/workflow/camunda/client.py
johandahlberg: No, I agree it might not be super relevant here. I just wanted to bring it since it is my experience that in any distributed system you will have these type of issues, even if it is all running on the same machine, etc. However, of course the larger the "distance" between systems, the higher the risk.

src/clims/services/workflow/__init__.py
withrocks: I'll remove it!

src/clims/services/workflow/__init__.py
withrocks: Right! I documented it in a strange place, in the demos. I'll move that doc.

src/clims/services/workflow/__init__.py
withrocks: Yes, I haven't looked into it as it was not needed to communicate with the camunda docker img. I'll add a ticket for revisiting this.

src/clims/services/workflow/__init__.py
withrocks: Right, I decided, perhaps without good reason, I don't know, to not lock unnecessarily down to using camunda. So the id is on the URL like format "camunda/GUID". This is just to keep the option open to use other workflow engines.

src/clims/services/workflow/__init__.py
withrocks: This is basically just an assert. It will raise an error if we are in a transaction (assert_no_transaction would be clearer!)

This is a trick I'm not sure we really require, because tests could also "ensure" that this doesn't happen.

The thing is that I have three sequential actions, (POSTGRES1, CAMUNDA, POSTGRES2). I want to ensure that if POSTGRES2 fails, I still have a record of POSTGRES1. If they are in a transaction, that would not happen, because both are rollbacked.

withrocks: Btw, I changed it to assert_no_transaction.

src/clims/services/workflow/__init__.py
withrocks: Yeah, we have a ticket on it.

src/clims/services/workflow/camunda/client.py
withrocks: Yes, I think it's a good point.

src/clims/api/endpoints/process_assignments.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import absolute_import
+
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+
+from sentry.api.base import Endpoint, SessionAuthentication
+
+
+class ProcessAssignmentsEndpoint(Endpoint):
```
GitEdvard, line 9: I feel a bit reluctant to the name "process". It's a super generic word and I don't know what's included in "process" in this context. The comments mention workflow. Why not WorkflowAssignment?

src/clims/api/serializers/models/task.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import absolute_import
+
+from rest_framework import serializers
+
+
+class TaskSerializer(serializers.Serializer):
+    id = serializers.CharField()
+    tracked_object = serializers.SerializerMethodField()
```
GitEdvard, line 8: I don't grasp what tracked object is here.

tests/clims/api/endpoints/test_task_definition.py
```diff
@@ -0,0 +1,42 @@
...
+
+        # Install with plugin API
+        sample_count = 2
+        cont = ExamplePlate(name="cont1")
+        for x in range(1, sample_count + 1):
+            sample = ExampleSample(name="sample-{}".format(x))
+            cont.append(sample)
+        cont.save()
+        workflow = SequenceSimple()
+        workflow.assign(cont)
```
GitEdvard, line 27: Shouldn't it be samples that are assigned to workflows?

tests/clims/api/endpoints/test_task_definition.py
```diff
@@ -0,0 +1,42 @@
...
+
+        # Get through REST API
+        self.login_as(self.user)
+        response = self.client.get(self.get_url())
+        expected = [{
+            "id": "clims.plugins.demo.dnaseq.workflows.sequence.SequenceSimple/data_entry",
+            "count": sample_count,
+            "name": "Data entry",
+            "processDefinitionKey": "clims.plugins.demo.dnaseq.workflows.sequence.SequenceSimple",
+            "processDefinitionName": "SequenceSimple",
```
GitEdvard, line 37: Hm, workflow and process seem to be synonyms here? In that case I would prefer the name workflow all over.

src/clims/models/assignment.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import absolute_import
+
+from django.db import models
+from sentry.db.models import (FlexibleForeignKey, sane_repr)
+from jsonfield import JSONField
+
+
+class SubstanceAssignment(models.Model):
+    """
+    Represents an assignment of a substance to a process.
```
GitEdvard, line 10: I dont like that process and workflow is used interchangably

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+            raise AssertionError("Expecting only one task name when creating a work batch, got: '{}'".format(name))
+        name = name[0]
+
+        # 2. Create a work batch with the "tracked objects" we just fetched
+        from clims.models.work_batch import WorkBatch
+        batch = WorkBatch(organization=organization, name=name)
+        batch.save()
+
+        for task in tasks:
+            task.tracked_object._archetype.work_batches.add(batch)
```
GitEdvard, line 285: I would rather have this encapsulated in tracked object. I have hard time to follow why _archetype is referensed, and what type archetype is.

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+    def get_handler(self, workflow_cls):
+        for handler_config in self.handlers.values():
+            configured_class = handler_config["cls"]
+            if issubclass(workflow_cls, configured_class):
+                return handler_config["handler"]
+            else:
+                raise NotImplementedError(
+                    "Workflow class '{}' has no handler".format(workflow_cls))
+
+    def AssignmentModel(self, item, user):
```
GitEdvard, line 315: Why isn't this method following python standard formatting?

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+        self._context.app.workflows.assign_item(self, item, self._context.user)
+
+    def assign_container(self, container):
+        if self.assign_substances_rather_than_containers:
+            for item in container.contents:
+                if isinstance(item, SubstanceBase):
+                    self.assign_item(item)
+        else:
+            self.assign_item(container)
+
```
GitEdvard, line 58: This gave me some thoughts. As I see it, we don't have this as explicit requirement from snpseq, but maybe it is a req from a future user of clims? I would like to see this type of functionality in a plugin, rather than in the core framework. This objection is more on a principal level. Each of these extra functionalities gives extra burden to the code.

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+            ret[formatted_name] = variables
+        except KeyError as e:
+            raise FormatError(
+                "Not able to interpolate {} into '{}'. Make sure that the name of the "
+                "preset refers to one of the variable names specified: {}".
+                format(six.text_type(e), name, kwargs.keys()))
+    return ret
+
+
+class FormatError(Exception):
```
GitEdvard, line 186: One idea, maybe we could gather exceptions and one line classes in a separate file? When looking at the _init_ file in collapsed view, all this mini-classes get the same 'weight' as the big classes. So that would get a more instanteounus overview.

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+        handler.start_workflow(workflow, item)
+
+        assignment.status = SubstanceAssignment.STATUS_DELIVERED
+        assignment.save()
+
+    def move_substance(self):
+        raise NotImplementedError()
+
+    # TODO: Naming
+    def batch_get_tracked_objects(self, category, keys):
```
GitEdvard, line 362: I have a hard time to grasp what tracked objects are. Perhaps this behavior (method) could be within a class rather than a service, and this class could tell in what context these tracked objects occur?

GitEdvard: I see that 'tracked_objects' is part of Task model (?). But for me, tracked objects would be something to find within a workflow? In clarity, a step is something that occur at (approximately) a single time. How does Task correlate to that? The existense of tracked objects tells that something is happening to the samples during a stretch of time.

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+class WorkflowService(object):
+    AssignmentError = AssignmentError
+
+    def __init__(self, app):
+
+        self._app = app
+
+        from .camunda import CamundaWorkflow
+        from .camunda.client import CamundaClient
+
```
GitEdvard, line 207: I would propose to have the camunda classes to a much higher degree isolated from the clims code. My vision is to have camunda classes in a separate repo actually, and then a translation layer to the clims repo.

GitEdvard: I think there is an opportunity to modularize here, since there is a natural border between Camunda and clims.

docs/development.md
```diff
@@ -72,6 +72,15 @@ You can get a fresh install of your database by running: `make reset-db`
 - Register the route to these endpoints in `sentry/api/urls.py`
 - Create serializers for the domain objects in `sentry/api/serializers/models/samples.py`
 
+# Adding framework tests
+
+Testing is currently almost only in form of integration tests, that at a minimum require postgres (and Django). The plan is to move these gradually to conform to the following:
+
+* Endpoint tests (tests/clims/api/endpoints) should be integration tests, requiring various middleware, e.g. postgres. These generally require no mocking but do require cleanup code or be run in transactions.
+* Serializer tests (tests/clims/api/serializers): Should be unit tests.
+* Service tests (tests/clims/services) are unit tests as we want to test the business logic within the service with mocked data. Queries to the db layer should be mocked, as well as any other calls to backends. Endpoint tests should cover accessing real data.
```
GitEdvard, line 81: I dont understand this. Above it's written that almost all tests are integration tests, which is true (meaning contacting a test db?). services has a lot of transaction methods in it, so they cant be unit tests?

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+            'process_definition__key').annotate(count=Count('name'))
+
+        for entry in task_definitions_with_instance_count:
+            entry["id"] = "{}/{}".format(
+                entry["process_definition__key"], entry["task_definition_key"])
+
+        return task_definitions_with_instance_count
+
+
+class ProcessTask(object):
```
GitEdvard, line 528: Why not just Task? ProcessTask feels confusing.

src/clims/services/workflow/camunda/client.py
```diff
@@ -0,0 +1,370 @@
...
+from clims import utils
+from clims.models import Workflow
+
+logger = logging.getLogger(__name__)
+
+
+class CamundaClient(object):
+    """
+    A client for executing workflows that interfaces with the underlying workflow engine.
+    It's higher level than CamundaApi which directly maps to the CamundaApi.
```
GitEdvard, line 20: This sentence was strange.

src/clims/services/workflow/__init__.py
GitEdvard: Workflows, and all in it, are both represented in Camunda and and in clims, but in different ways. So the internal structure of these classes will likely be different as well. I fear that there will be classes that are partly the same, but subtly different anyway, and that will be confusing.

GitEdvard: My vision here is that this translation layer should be thin. And this is the only place that Camunda classes and clims classes co-exists.

GitEdvard: There is also a linguistic boundary between Camunda and clims. Camunda lingo is not set by ourself, so we have to adapt to it.

GitEdvard (commented):
Hello! I decided to send in reviews in a few separate steps, and this is my first. It get to big otherwise. Enjoy my feedback :-)

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+            else:
+                # The subclass has defined presets as a dict:
+                return self.presets
+
+        # No definitions found
+        return dict()
+
+
+def create_preset(name, **kwargs):
+    """
```
GitEdvard, line 132: I feel that I miss a visual representation on e.g. a whiteboard on how these classes are thought to relate to each other. I know that you have made a demo on it and all, but right now I'm kind of lost :(. I feel that this hinders me right now to give meaningful response.

GitEdvard: This is my guess right now, a Workflow consists of tasks. Workflow can be visualized in Camunda (and in clims?). A preset is a workflow that is pre-configured.

GitEdvard: The concept of Protocol in Clarity is skipped in clims?

src/clims/api/endpoints/process_assignments.py
withrocks: Yes I understand that, thanks for bringing this up. :)

(Please note first that this is in the middle of refactoring it to always be named process, never workflow, that explains the chaotic switches between names.)

This is a bit of a long explanation of my thinking here, as it also goes into how I suggest we think of naming in general. It would be great to get a consensus on what's acceptable in the project so I'd like to ping the team too: @commonlims/core.

**Workflow vs. Process**

So why not workflow instead of process? It's a tough decision in my mind, because in Camunda and in the BPMN standard (check out https://www.omg.org/spec/BPMN/2.0/PDF, section Process) these are called processes. They sometimes call it Workflow, but informally. The executable workflow is called a Process.

If we start naming it differently there is a certain disconnect between what we call it and what Camunda calls it, which complicates things and creates work for us, because we need to create new names for some things (e.g. `process_definition_key` must be `workflow_definition_key`). Also, when debugging you have to remind yourself of this mapping all the time.

I thought it might be a good middle ground to call it "workflows" at the top layer, but "process" when communicating with Camunda, and this is what I started doing. But that also has the downside in that it gets confusing when you get further down the stack and everything is all of a sudden called process.

I have a tendency to think, having worked a bunch with the workflow feature for a while now, that it's simpler to just call it the same as Camunda calls it and then document clearly what processes are in the context of Common LIMS.

**Naming in general**

I think it's fine to use generic words in programming, as long as it's clear to everyone what the generic word means in the scope. The team discussed a bunch until we came up with the word `Substance` for what's in a container, and I think that was great, but it's definitely still a generic word. Problem is that all other words we thought of were also generic, only some more than others (e.g. Item).

An example of a highly generic word we use regularly is when we use the the word "print" to print to the console in Python. It's a generic word in the sense that it can also mean "print to a printer" (it even means that in javascript), but every Python programmer knows that it doesn't mean that in the context of a Python program. I would argue that it's OK that they thought it meant "print to a printer" their first day learning Python, as long as they know better the second day :)

**Conclusion**

So from my perspective it's OK to call workflows processes, because we're in the context of a workflow engine. Also, because it's "process assignment" it's hopefully clear that it must be assignment to a workflow process, because we don't have the concept of assigning to e.g. an operating system process. And this will in the end be clear in our documentation.

**Follow up**

I suggest a guideline which the whole team (ping @commonlims/core) might want to chime in on: If we think that a term becomes confusing after reading the documentation (of the end product), we should change it, otherwise it's ok.

And of course, if you think otherwise, please say so!

src/clims/services/workflow/__init__.py
GitEdvard: One thought here, the lab work is divided between two groups in the lab, pre and post. Is there any way to visualize which of the samples are in the pre-lab work, and which are in the post-lab work?

GitEdvard: This might be something to bring up to the lab, how they feel about it?

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+    value in that iterable. If there are more than one iterable, the product of them is used.
+
+    Example:
+
+    >>> create_preset("{sequencer} {method}", sequencer=["A", "B"], method=["m1", "m2"], sample_type="DNA")
+    {
+      'A m1': {'method': 'm1', 'sequencer': 'A', 'sample_type': 'DNA'},
+      'A m2': {'method': 'm2', 'sequencer': 'A', 'sample_type': 'DNA'},
+      'B m1': {'method': 'm1', 'sequencer': 'B', 'sample_type': 'DNA'},
+      'B m2': {'method': 'm2', 'sequencer': 'B', 'sample_type': 'DNA'}
```
GitEdvard, line 147: Is it so that there is a set of tasks that comes with a certain combination of sequencer, method and sample type?

src/clims/api/serializers/models/task.py
withrocks: Right. This is something we will have to document.

Camunda has the concept of adding a "business key" to a process. When you start a process, you can assign a business key to it, which is just a process variable, but it's a special one, in that you can query for it specifically and I assume it's optimized to query for it in some cases.

The idea is that this "business key" is what the process instance is handling. I came up with the term `tracked_object` to refer to this object, as it's tracked through the whole workflow.

In our context, the `tracked_object` is always a `Substance` currently, but in the future it might for example be a `Container`. If it's a substance, the business key of the tracked object will be `Subtance/<id in our database>`.

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+    def create_work_batch(self, task_ids, organization):
+        """
+        Creates a new WorkBatch for a list of tasks
+        """
+
+        # 1. Fetch all the tasks from the task IDs we received
+        tasks = self.get_tasks_by_ids(task_ids)
+        name = list({t.name for t in tasks})
+        if len(name) > 1:
+            raise AssertionError("Expecting only one task name when creating a work batch, got: '{}'".format(name))
```
GitEdvard, line 276: This was cryptic for me. Why does the client send in an array of task ids, and then it's understood that all these should have the same name?

tests/clims/api/endpoints/test_task_definition.py
withrocks: Yes, but we also (currently at least) support to assign a whole container in the high-level API. The workflow `SequenceSimple` is configured to take each substance from the container and assign it to the workflow. I figured that this might be handy, but maybe it's just confusing.

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+                .format(installed.version, db_version))
+
+    @transaction.atomic
+    def create_work_batch(self, task_ids, organization):
+        """
+        Creates a new WorkBatch for a list of tasks
+        """
+
+        # 1. Fetch all the tasks from the task IDs we received
+        tasks = self.get_tasks_by_ids(task_ids)
```
GitEdvard, line 273: I miss an explicit class for Task. When I follow the reference (get-tasks-by-ids) I have still hard time to decipher what a Task is. It would be easier for me if there was just a dedicated place for Task.

src/clims/services/workflow/__init__.py
```diff
@@ -0,0 +1,546 @@
...
+        name = list({t.name for t in tasks})
+        if len(name) > 1:
+            raise AssertionError("Expecting only one task name when creating a work batch, got: '{}'".format(name))
+        name = name[0]
+
+        # 2. Create a work batch with the "tracked objects" we just fetched
+        from clims.models.work_batch import WorkBatch
+        batch = WorkBatch(organization=organization, name=name)
```
GitEdvard, line 281: Workbatch seem to be a like a step in Clarity, and tasks is like the different scripts within a step? Is a work batch performed by one person in the lab?

GitEdvard: If it is as I guessed above, maybe these classes should be named as Task and SubTasks?

src/clims/services/workflow/__init__.py
GitEdvard: Perhaps this method could reside within the substance and container classes respectively?

src/clims/services/workflow/__init__.py
GitEdvard: I dont follow you here. Why are there two postgres instances? I also thought that camunda should be in postgres, so that it should be only one db instance?

GitEdvard: I think the complexity will rise with more than one database to keep track on. Is there a problem with having one db?

GitEdvard: (these comments refer to the previous discussion between Johan and Steinar, and I hope that it will be clear when posting my review... )

GitEdvard (commented):
Here comes my second review. Note that we have not finished discussing the first review, so some things may already have been resolved.

PR #139, opened 2020-03-25
New ports for postgres

withrocks:
Using a different port than 5432 since it may clash with an already
installed version.


GitEdvard:
Super, make fresh works for me now! But, there is conflicts. 

Note that the change in clims.conf.py.default, is not effective if you already have a clims.conf.py file in ~/.clims folder. So you have to write in that change manually to make it work. (Maybe it works if you remove that file?)

withrocks:
Cool! But yeah, the easiest thing is to just remove ~/.clims and then run `lims init` to force a new config file. I'll check out the conflict.

PR #142, opened 2020-06-12
278, expandable list, group samples on containers

GitEdvard:
Purpose:
Continue with the work on showing an expandable/grouped list at the "Samples" entry point. Here, grouped headers are shown for both "sample type" and "containers" in the "group by" dropdown list. However, the *expand* functionality is not yet implemented, i.e. pressing the downarrow at the margin. 

Refactorings:
Move some logic from the redux action/reducer classes to a utility module. Also, assert that the reducer don't mutate state. 

Shortcommings:
As implemented now, both sample ids and container ids is used in the array byIds in substanceSearchEntry. There will probably be clashes between them in future. I have written a ticket for this.

### DIFF
--- a/src/sentry/static/sentry/app/components/listView.jsx
+++ b/src/sentry/static/sentry/app/components/listView.jsx
@@ -99,6 +99,14 @@ class ListView extends React.Component {
     }
   }
 
+  getFontStyle(entryId, header) {
+    const row = this.props.dataById[entryId];
+    if (header.fontstyle !== undefined) {
+      return header.fontstyle(row);
+    }
+    return 'normal';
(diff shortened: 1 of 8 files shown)


withrocks:
Looks good! Note that I have a MAJOR refactoring of all reducers and actions coming up. I'll go through them in detail with you after merging this stuff with it.

withrocks:
... and yeah, I apologise if that one will cause work for you, I intend for it not to. But the reason for that refactoring is that we'll have to remove all the duplication of code that the different resources now have. I hope that if we just communicate a lot it won't be too much of an issue.

PR #148, opened 2020-06-16
Add the reselect library

withrocks:
Introduces the `reselect` library. It's designed to create memoized selectors on top of react data. 

Recommended reading: https://redux.js.org/recipes/computing-derived-data

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "redux-mock-store": "^1.5.3",
     "redux-thunk": "^2.3.0",
     "reflux": "0.4.1",
+    "reselect": "^4.0.0",
     "scroll-to-element": "^2.0.0",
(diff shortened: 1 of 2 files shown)


GitEdvard:
It looks nice. I'm concerned however about the possible increased complexity this might lead to, and that we then loose momentum to implement the basic functionality of commonlims?

withrocks:
I understand your concern. I think however that this doesn't have to affect anything if one doesn't want to (to begin with at least). This makes some things much cleaner I've found - and it's a recommended approach to certain problems by the core redux team.

This PR doesn't have the context as I'm breaking things down a lot, but I brought it in because it's required (or something similar) to fetch process presets out of a list of process definitions. I hope it will be clearer when you see that PR.

GitEdvard:
Ok!

PR #149, opened 2020-06-16
Add helper methods for creating shared reducers and actions

withrocks:
A lot of our actions and reducers duplicate code as it stands today.
This introduces methods for generating both of these with factory
functions.

### DIFF
--- a/src/sentry/static/sentry/app/redux/actions/shared.js
+++ b/src/sentry/static/sentry/app/redux/actions/shared.js
@@ -2,7 +2,7 @@ import axios from 'axios';
 
 // Helper function that creates an action creator.
 // From: https://redux.js.org/recipes/reducing-boilerplate
-export function ac(type, ...argNames) {
+export function makeActionCreator(type, ...argNames) {
   return function(...args) {
     const action = {type};
     argNames.forEach((arg, index) => {
@@ -80,7 +80,7 @@ export const acSelect = resource => {
   };
 };
 
-export const list = {
+export const resourceActionCreators = {
   acRequest,
   acSuccess,
   acFailure,
@@ -93,13 +93,13 @@ export const list = {
 export const makeResourceActions = (resourceName, getListUrl) => {
   return {
     // Fetching a list
-    getListRequest: list.acRequest(resourceName),
-    getListSuccess: list.acSuccess(resourceName),
-    getListFailure: list.acFailure(resourceName),
-    getList: list.acGetList(resourceName, getListUrl),
+    getListRequest: acRequest(resourceName),
+    getListSuccess: acSuccess(resourceName),
+    getListFailure: acFailure(resourceName),
+    getList: acGetList(resourceName, getListUrl),
 
     // Select entries in the list or pages of it
-    select: list.acSelect(resourceName),
-    selectPage: list.acSelectPage(resourceName),
+    select: acSelect(resourceName),
+    selectPage: acSelectPage(resourceName),
   };
 };
--- a/src/sentry/static/sentry/app/redux/actions/shared.js
+++ b/src/sentry/static/sentry/app/redux/actions/shared.js
@@ -1,3 +1,5 @@
+import axios from 'axios';
+
 // Helper function that creates an action creator.
 // From: https://redux.js.org/recipes/reducing-boilerplate
 export function ac(type, ...argNames) {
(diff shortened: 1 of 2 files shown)


src/sentry/static/sentry/app/redux/actions/shared.js
```diff
@@ -9,3 +11,95 @@ export function ac(type, ...argNames) {
...
+    };
+  };
+};
+
+export const acSuccess = resource => {
+  return (entries, link) => {
+    return {
+      type: `GET_${resource}_LIST_SUCCESS`,
+      entries,
+      link,
```
GitEdvard, line 34: When it comes to substance search entries, they have "groupBy" as well. How about including that?

src/sentry/static/sentry/app/redux/actions/shared.js
```diff
@@ -0,0 +1,11 @@
+// Helper function that creates an action creator.
+// From: https://redux.js.org/recipes/reducing-boilerplate
+export function ac(type, ...argNames) {
```
GitEdvard: I don't agree that "ac" should add to the readability over "makeActionCreator". I prefer the longer one. Often when I'm new to a code, or revisit it, I sometimes have a hard time to decipher abbreviations.

src/sentry/static/sentry/app/redux/actions/shared.js
```diff
@@ -9,3 +11,95 @@ export function ac(type, ...argNames) {
...
+  return (id, doSelect) => {
+    return {
+      type: `TOGGLE_SELECT_${resource}`,
+      id,
+      doSelect,
+    };
+  };
+};
+
+export const list = {
```
GitEdvard: Perhaps "actionCreators"

src/sentry/static/sentry/app/redux/reducers/shared.js
```diff
@@ -0,0 +1,219 @@
...
+// Returns an altered listViewState where a single element has been selected or deselected
+// If select is null, the item will be toggled. If it's true, it's selected, if it's false
+// it's deselected
+export function selectSingle(state, action) {
+  const {listViewState} = state;
+  const {id, doSelect} = action;
+
+  const shouldToggle = doSelect === null || doSelect === undefined;
+  const shouldSelect = shouldToggle ? !listViewState.selectedIds.has(id) : doSelect;
+  const newSet = shouldSelect
```
GitEdvard: Just selectedIds instead of newSet?

GitEdvard (approved):

src/sentry/static/sentry/app/redux/actions/shared.js
withrocks: Yes, I was thinking of doing that when we start using this for substance search entries.

src/sentry/static/sentry/app/redux/actions/shared.js
withrocks: I agree, and also kind of don't :) In this particular case I feel that the short name was a good choice because `ac` is being used as almost a metaprogramming device, as it were part of the programming language - and only used in the definition of actions. It's similar to the function `t` that's used all over the place in the javascript code for translated strings. So I liked this because when you read the action files, it will be more readable, as this is so easy to read:
```
getCustomer = ac('GET_CUSTOMER', 'name', 'age')
getCustomers = ac('GET_CUSTOMERS') 
```

But, I also thought that if there would be any objections, I would gladly go back to having it the other way. I think I'll change it back in another PR if it's fine with you.

src/sentry/static/sentry/app/redux/actions/shared.js
withrocks: Ahh, you mean `listActionCreators` or? Because the idea is that these are generators for `list` actions. The idea is not that this is the list of all action creators.

But since I needed to explain that, perhaps `listActionCreators` would be better.

src/sentry/static/sentry/app/redux/reducers/shared.js
withrocks: Yes, that's a good point!

src/sentry/static/sentry/app/redux/actions/shared.js
withrocks: I'll change it now, since I have another thing to change.

src/sentry/static/sentry/app/redux/actions/shared.js
GitEdvard: ok

src/sentry/static/sentry/app/redux/actions/shared.js
GitEdvard: :+1:

PR #152, opened 2020-06-25
Refactor shared action/reducer helpers

withrocks:


johandahlberg (approved):
Very nice. My only comment is that it was not completely obvious to me that `ac=ActionCreator`. I don't know if you want to spell that out, or possibly remove it and instead make it obvious what type of thing one is dealing with based on the context from which it is exported?

johandahlberg:
That said, the above is just a thought, and I'll go ahead and merge this now.

withrocks:
I agree, now that I think of it it could just as well mean action, so spelling out actionCreator would probably be a good idea! Will keep it in mind.

PR #153, opened 2020-06-25
Allow suppressed warnings in test output

withrocks:
We have had a giant block of warnings because of renaming in React. We
do have a ticket on resolving that, and it's not finished because
upgrading emotion to 10 is not entirely trivial.


johandahlberg:
lgtm

PR #154, opened 2020-07-23
333 expand sample type

GitEdvard:
Purpose:
To be able to expand/collapse list view entries when grouped. This PR specificly handles substances that are grouped by sample type. 

Implementation:
Add a couple of expand/collapse events into the substanceSearchEntry action/reducer. Expanded entries are cached, so that the second time the same entry is expanded, the cache is used. 

Introduce the concept of entity and entry in substanceSearchEntry. Here. entity = raw api response, and entry = entity plus metadata, meaning for example children.isExpanded or children.cachedIds.

### DIFF
--- a/src/clims/services/base_extensible_service.py
+++ b/src/clims/services/base_extensible_service.py
@@ -190,6 +190,9 @@ class BaseExtensibleService(object):
 
 
 class BaseQueryBuilder:
+    # TODO: decide internal query separators
+    QUERY_SEPARATOR = "&&"
+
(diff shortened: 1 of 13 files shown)


withrocks:
Rebased and removed that commit that remove the vscode files. Locally tested (as the github actions have an unresolved issue.)

withrocks:
Rebased and removed that commit that remove the vscode files.

withrocks:
Tests working locally.

PR #155, opened 2020-07-30
Harden the linting process

withrocks:


yarn.lock
```diff
@@ -10677,15 +10677,15 @@ prettier-linter-helpers@^1.0.0:
...
-  version "1.15.3"
-  resolved "https://registry.yarnpkg.com/prettier/-/prettier-1.15.3.tgz#1feaac5bdd181237b54dbe65d874e02a1472786a"
-
-prettier@^2.0.5:
+prettier@2.0.5:
   version "2.0.5"
   resolved "https://registry.yarnpkg.com/prettier/-/prettier-2.0.5.tgz#d6d56282455243f2f92cc1716692c08aa31522d4"
   integrity sha512-7PtVymN48hGcO4fGjybyBSIWDsLU4H4XlvOHfq91pz9kkGlonzwTfYkaIEwiRg/dAJF9YlbsduBAgtYLi+8cFg==
 
+prettier@^1.14.3:
```
GitEdvard, line 10685: I don't understand why there's two entries for prettier with different versions?

.vscode/settings.json
```diff
@@ -26,11 +26,13 @@
   },
 
   "[javascript]": {
-    "editor.formatOnSave": true
+    "editor.formatOnSave": true,
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
GitEdvard, line 30: Is it possible to set the prettier version in config file as well?

docs/development/code-quality.md
```diff
@@ -0,0 +1,27 @@
...
+## Javascript linting
+
+The javascript code is linted using `eslint`. Relevant configuration files are:
+
+Explicit command: `make lint-js`
+
+* .eslintignore: Files to be ignored during linting
+* .eslintrc.js
+
+## Javascript formatting
```
GitEdvard, line 23: Perhaps we should write explicitly what's included in linting? I for my self is not so sure what's included other than formatting and unit tests.

yarn.lock
withrocks: This has almost certainly to do with that some other package requires prettier of a certain version. Node is good about that kind of version management. I generally never review the yarn.lock file because it has a ton of information like this which is hard to read and is automatically generated when we run yarn install. The file package.json is the one it bases all of its resolutions on.

.vscode/settings.json
withrocks: This plugin uses the correct version of prettier (the one configured in package.json) if it's installed. So if `./bin/lint --check-deps-only` runs without an issue it should use that version.

docs/development/code-quality.md
withrocks: Yes good point, I hope to get back to this when I have time :) For now, I hope that the image that I added (but forgot to link to in this document) gives enough information, because it says which configuration files are being used. I'll add a link to it now.

yarn.lock
GitEdvard: OK!

.vscode/settings.json
GitEdvard: nice!

docs/development/code-quality.md
GitEdvard: good enough!

withrocks:
Locally validated.

PR #156, opened 2020-07-31
322 merge location

GitEdvard:
Purpose:
Get a nicer representation of container index, so that e.g. plate wells are viewed as "A:3", instead of "(0,2,0)". Also merge the two internal properties "location" and "container index" on substance. Location was previously returning a django object, now it's returning the domain object of type ContainerIndex.

Points of careful concern:
Rendering a plugin extensible class that is not registered will now cast exception instead of silently fallback to SubstanceBase. This change was "inherited" from the work Steinar has done in March. I think this was what we agreed on back then.


withrocks (approved):

withrocks:
Locally tested (github actions are down). Merging.

PR #157, opened 2020-08-04
377 trigger plugin script

GitEdvard:
Purpose:
To give a proposal how to trigger plugin scripts. This should be configured by a technical user in an easy manner, by creating a class inherited from Step.

Scenario:
1. User selects samples and enters a step. Here there is an api call to give a list of button names for the UI to show.
2. User is within a step and presses a button to trigger a script. Here there is another api call to trigger the script linked to the button by means of the user configuration file.

The class MyFancyStep in test_step_configuration.py is an example of an user defined configuration that should reside within plugin code. 

Implementation:
The "convention" that was mentioned in the Jira ticket, how to link together an event in UI with a plugin script, is here implemented by means of decorators. I wanted to give the option to give a button text, which required an argument that don't belong to the actual method call. 

Considerations:
I think it would be handy for the user to define step-properties (step-udfs) in the same configuration file. This made me think that maybe these configuration files should inherit from the workbatch class, which in turn  inherits from extensible. What do you think of that? I am a bit hesitant to that idea though. I would like to have step-properties in the config file, but the config file separated from instances of workbatch.


src/clims/api/endpoints/trigger_script.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class TriggerScriptEndpoint(OrganizationEndpoint):
+    authentication_classes = DEFAULT_AUTHENTICATION
+    permission_classes = (IsAuthenticated, )
+
+    def post(self, request, organization):
+        # This endpoint is constructed in order to trigger a plugin script,
+        # specified by input parameters.
+        serializer = ScriptTriggerSerializer(data=request.data)
```
withrocks, line 19: My idea was that the UI programmer can never trigger a script directly, so in that case an endpoint would not be required for this. Rather I envisioned it so that the custom scripts would always be triggered by some service class.

Example:
* User creates a workbatch via the WorkflowService
* The workflow service triggers the on_create handler

withrocks:
I'm adding some comments here that will be clearer when my next PRs (and diagrams) will be committed.

src/clims/configuration/hooks.py
```diff
@@ -0,0 +1,13 @@
+from __future__ import absolute_import, print_function
+
+HOOK_TYPE = '_hook_type'
+
+HOOK_TAG = '_hook_tag'
+
+
+def button(button_text):
```
withrocks, line 8: I think the idea of having a decorator for this is the way to go, as we'll both have a way to describe the UI as well as handling the event in the same place, which follows the api-simplicity design goal we have :+1:

src/clims/models/step.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import absolute_import, print_function
+
+from django.db import models
+from sentry.db.models import (FlexibleForeignKey, sane_repr)
+from clims.models.extensible import ExtensibleModel, ExtensibleVersion
+
+
+class Step(ExtensibleModel):
```
withrocks, line 8: I wonder still if we need a django model for this or not. We already have a django model that says that version x of a certain plugin is installed, so in that way we have traceability for this. We might want to discuss this via a chat.

src/clims/services/application.py
```diff
@@ -24,6 +25,7 @@ def __init__(self):
         from sentry.plugins import PluginManager
         self.plugins = PluginManager(self)
         self.workflows = WorkflowService(self)
+        self.workbatches = WorkbatchService(self)
```
withrocks, line 28: In the PR I'll be presenting soon, the WorkBatches are all handled in the WorkflowService. I think that that makes sense because they are parts of the workflow. We can merge the two later.

src/sentry/plugins/base/manager.py
```diff
@@ -117,6 +118,16 @@ def install_plugin(self, plugin_cls):
...
 
+    def install_step_templates(self, plugin):
+        """
+        Installs all the step tempalates found in the plugin.
+        """
+        logger.info("Installing step templates found in plugin class '{}'".format(
+            plugin.get_name_and_version()))
+
+        for step_template_cls in plugin.get_step_templates():
+            self._app.workbatches.register_step_template(step_template_cls)
```
withrocks, line 129: I think another way to approach this is to have it in-memory only. That relates to not having models for the steps. I think there are good arguments for both approaches - currently we're doing a blend of both. I hope the diagrams will be able to shed light on how it's all today and then we have a better ground to decide which way we should go (pros/cons for each approach).

withrocks:
Looks cool! I'll give more feedback when my PR is ready. Then we can discuss this with diagrams!

src/clims/api/endpoints/trigger_script.py
GitEdvard: I don't follow you here. I don't see how an api endpoint count NOT be triggered after a user buttonclick?

src/clims/configuration/hooks.py
GitEdvard: :+1:

src/clims/models/step.py
GitEdvard: Yes. This inheritance was kind of a leftover since I was experimenting back and forth. I'll remove this one.

src/clims/services/application.py
GitEdvard: That sounds good for me. At this stage of work, I just needed a place to place my code.

src/sentry/plugins/base/manager.py
GitEdvard: Ah, yes. This code was also something I temporarily put up in order to make it work for now. In fact, the Step extensible model (which I forgot to remove at code where the previous comment were), is actually never registered into db (the code differ from other sorts of registrations). However, in the end I think it would be sound to have an automatic validation that there are no two steps with the same name, which can easily be checked by putting the definitions into the db.

src/clims/api/endpoints/trigger_script.py
withrocks: Right, sorry for the confusion, we'll need something like this in the case of button clicks, but in the case of `on_create` I was thinking that it happens when the workbatch gets created.

About the name "trigger script". I wonder if we could have it something like "trigger action"?

withrocks (approved):

PR #160, opened 2020-08-09
Minor changes

withrocks:
* Use the binary version of postgres so it doesn't need to be built during pip install
* Allow the list view to not require items with a group header (backwards compatibility)
* Remove the loading triangle

### DIFF
--- a/src/sentry/static/sentry/app/components/listView.jsx
+++ b/src/sentry/static/sentry/app/components/listView.jsx
@@ -109,7 +109,11 @@ class ListView extends React.Component {
   getDisplayCell(entryId, header) {
     const row = this.props.dataById[entryId];
     if (typeof header.accessor === 'function') {
-      return header.accessor(row.entity, row.isGroupHeader);
+      if (row.isGroupHeader !== undefined) {
+        return header.accessor(row.entity, row.isGroupHeader);
+      } else {
+        return header.accessor(row);
+      }
     }
     return row[header.accessor];
   }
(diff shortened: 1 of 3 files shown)


withrocks:
In case the integration tests time out again, I've tested this successfully locally.

src/sentry/static/sentry/app/components/listView.jsx
```diff
@@ -109,7 +109,11 @@ class ListView extends React.Component {
   getDisplayCell(entryId, header) {
     const row = this.props.dataById[entryId];
     if (typeof header.accessor === 'function') {
-      return header.accessor(row.entity, row.isGroupHeader);
+      if (row.isGroupHeader !== undefined) {
+        return header.accessor(row.entity, row.isGroupHeader);
+      } else {
+        return header.accessor(row);
```
GitEdvard, line 115: In the first case here, the api response is wrapped into the "entity" attribute (to distinguish from UI metadata), where as the other case, the api response is taken as is and named row. I'm not so delighted to permit two different json structures for the "row" variable. Do we have any backwards compatibility to respect here?

GitEdvard (commented):
Ok, I have one comment.

src/sentry/static/sentry/app/components/listView.jsx
withrocks: No I agree that this isn't optimal! I had another view that used the list with data that wasn't formatted like this. Out of the two options of changing the data to have this "isGroupHeader" and changing the view, I did a quick change to the view, thinking that we could synch it in our tech debt rounds.

withrocks:
Do you mind if we merge this for now and make it optimal later?

PR #162, opened 2020-08-09
Make list action bar reusable

withrocks:


withrocks:
Validated locally.

src/sentry/static/sentry/app/views/projects/createProjectButton.jsx
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  render() {
+    return (
+      <React.Fragment>
+        <a
+          title={this.props.tooltip || this.props.buttonTitle}
+          className="btn btn-sm btn-default"
+          disabled={this.props.disabled}
+          onClick={this.onToggle}
```
GitEdvard, line 46: I'm expecting to see some create project form to show here? How is it that a create project button is linked to onToggle event?

src/sentry/static/sentry/app/views/substances/substances.jsx
```diff
@@ -139,12 +140,21 @@ class Substances extends React.Component {
...
     return (
-      <ListActionBar
-        realtimeActive={false}
-        query=""
-        orgId={orgId}
-        canAssignToWorkflow={canAssignToWorkflow}
-      />
+      <ListActionBar>
+        <div className="btn-group">
+          <UploadSubstancesButton className="btn btn-sm btn-default" disabled={false}>
```
GitEdvard, line 145: Cool, making it reusable!

GitEdvard (approved):

src/sentry/static/sentry/app/views/projects/createProjectButton.jsx
withrocks: Right! This is copied from e.g. the `AssignToWorkflowButton`. I think the naming is confusing, a this is both a view for the button and the modal window. Should rename to `Modal` or something (copied this naming convention blindly from sentry).

The onToggle function toggles between showing and not showing, so when clicking the button it leads to a toggle.

PR #164, opened 2020-08-10
TaskDefinition resource

withrocks:


tests/js/spec/redux/reducers/taskDefinition.spec.js
```diff
@@ -0,0 +1,68 @@
...
+  it('has expected state after requesting getting a list', () => {
+    const requested = taskDefinition(
+      initialState,
+      taskDefinitionActions.getListRequest()
+    );
+    const expected = {
+      loading: true,
+      errorMessage: null,
+      byIds: {},
+      listViewState: {
```
GitEdvard, line 15: This is not how the substanceSearchEntry state is organized. Should be write a ticket to make it consistent?

src/sentry/static/sentry/app/propTypes/resources.jsx
```diff
@@ -0,0 +1,16 @@
+import PropTypes from 'prop-types';
+
+// TODO: Auto generate this file from Python serializers
+
+export const TaskDefinition = PropTypes.shape({
+  id: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
+  count: PropTypes.number.isRequired,
+  processDefinitionName: PropTypes.string.isRequired,
```
GitEdvard, line 9: It seems to me that TaskDefinition is not the right description of the object used here. As I understand it, a TaskDefinition could be included by more than one process/workflow, and here there is a reference to one single process/workflow.

tests/js/spec/redux/reducers/taskDefinition.spec.js
withrocks: Right! I think we should have a ticket on using the shared actions and reducers for all shared resources. That would mean a lot less code and some tests we can get rid of (as they really test the same thing).

One thing that's missing in the shared reducers/actions as of now is that they don't take into account the parent/child relationship you worked on (which relates back to your other comment on my hack for showing things that don't have `isGroupHeader`).

It would be nice if that task would also merge that logic into the shared action/reducers:

Added: https://snpseq.atlassian.net/browse/CLIMS-392

src/sentry/static/sentry/app/propTypes/resources.jsx
withrocks: This is admittedly a little bit hairy at the moment, but this corresponds to a "UserTask" in a process diagram, which is bound to a particular process definition.

I think we should resolve this together on a meeting in september (https://snpseq.atlassian.net/browse/CLIMS-393)

GitEdvard: Why not simply rename it to "UserTask" then?

withrocks: I think that the API in Camunda calls it a task, if I remember correctly :D Naming is hard (tm) I think we need to go through all these naming issues together as a group and synch all of it. At the moment I'm afraid that any change will lead to another discussion when Mariya is back, and we change back to something else.

PR #165, opened 2020-08-11
389, Fix plate reference by container index

GitEdvard:
Purpose: 
Fix this kind of plate referensing:
plate_position = new PlateIndex(myplate, (0,1,0))
mycontainer[plate_position] = mysample

Points for extra concern:
I'm a bit worried that the container is part of container_index, not just the actual position. Suppose a use case: application programmer expects two plates where samples are placed identical. He wants to loop one plate and fetch corresponding contents in the other plate. So he uses plate index generated from the 1st plate, and fetch contents with plate2[plate_index]. Nothing wrong so far, but I smell bugs coming here. 

What do you think?

### DIFF
--- a/src/clims/services/container.py
+++ b/src/clims/services/container.py
@@ -52,6 +52,10 @@ class ContainerIndex(object):
             return cls.from_string(container, key)
         elif isinstance(key, tuple):
             return cls(container, *key)
+        elif issubclass(type(key), ContainerIndex):
+            # This has to be refactored in case there is problem
+            assert container == key.container, "containers are not matching in container index"
+            return key
         else:
             raise NotImplementedError("Can't use {} as an index".format(type(key)))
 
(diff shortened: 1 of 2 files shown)


withrocks:
I totally see that point now when you mention it. We might want to just ensure that the "bound" container on the plate_index is never used for anything but the UI. But I assume that there is a better approach than what I did there - I suggest we look further into it as we progress.

src/clims/services/container.py
```diff
@@ -52,6 +52,10 @@ def from_any_type(cls, container, key):
             return cls.from_string(container, key)
         elif isinstance(key, tuple):
             return cls(container, *key)
+        elif issubclass(type(key), ContainerIndex):
+            # This has to be refactored in case there is problem
+            assert container == key.container, "containers are not matching in container index"
```
withrocks, line 57: Good catch. Another option here is just to ignore the container lying on the index and act as it's not there. A starting point would be to make it private-looking so that it will definitely only be used by underlying code. But for now I think this assert is good so it reminds us we need to finalize this design.

withrocks (approved):

PR #166, opened 2020-08-13
Add get single to the shared redux actions

withrocks:


tests/js/spec/redux/actions/shared.spec.js
```diff
@@ -60,6 +60,32 @@ describe('shared action creators', () => {
       id: 'id',
     });
   });
+
+  it('should create get request', () => {
+    const action = resourceActionCreators.acGetRequest(RESOURCE_NAME)('id');
```
GitEdvard, line 65: This seem to me that "id" is the allowed parameter to enter here?

tests/js/spec/redux/actions/shared.spec.js
```diff
@@ -111,4 +137,38 @@ describe('shared async actions', () => {
       expect(moxios.requests.count()).toEqual(1);
     });
   }, 500);
+
+  it('can get a single entry', () => {
+    const store = mockStore({});
+    const url = '/url?search=search&cursor=cursor';
```
GitEdvard, line 143: Whouldn't it be more informative that the url should reflect that "id" is a required parameter?

tests/js/spec/redux/reducers/shared.spec.js
```diff
@@ -0,0 +1,234 @@
...
+  });
+
+  it('has expected state when getting a successful list response', () => {
+    const successState = createGetListSuccessState();
+    expect(successState).toEqual({
+      loadingDetails: false,
+      detailsId: null,
+      loading: false,
+      errorMessage: null,
+      byIds: {'1': {id: '1', name: 'entry1'}, '2': {id: '2', name: 'entry2'}},
```
GitEdvard, line 65: When coming to expandable lists, there is a lot of metadata containing UI state, such as isFetched, cachedIds, etc. The api entries are therefore embedded into "entity" attribute, to be distinguished from ui metadata. This make the shared reducers not compatible with expandable list reducers. I thought you should know that.

tests/js/spec/redux/reducers/shared.spec.js
```diff
@@ -0,0 +1,234 @@
...
+    const requestedState = reducer(initialState, actions.getRequest());
+    // We must get an item with an id back:
+    const fetchedItem = {
+      id: 1,
+    };
+    const successState = reducer(requestedState, actions.getSuccess(fetchedItem));
+    expect(successState).toEqual({
+      ...initialState,
+      detailsId: 1,
+      byIds: {1: {id: 1}},
```
GitEdvard, line 203: Shouldn't visibleIds be updated as well?

tests/js/spec/redux/actions/shared.spec.js
withrocks: In this case the string 'id' is the argument for the parameter with the same name. It's being sent to the function that acGetRequest creates dynamically. Does that answer the question?

tests/js/spec/redux/actions/shared.spec.js
withrocks: Yeah I believe you're right. In this case, the url can just be whatever because I only want to see if it was called, so I just set it to "url".

tests/js/spec/redux/reducers/shared.spec.js
withrocks: Right, thanks. I think it will be some work to merge the two. I've created a ticket on it.

tests/js/spec/redux/reducers/shared.spec.js
withrocks: Good question, but I don't believe it should. When one executes a search or similar one gets two things:

* A partial list (page) of data that's cached in byIds. Currently, byIds will be cleared, but in the future we might selectively keep it for faster paging back and forth or something like that.
* A refreshed list of visible IDs (because we're viewing it in the list view). (An argument could be made to have that view state entirely in the view components themselves, but at the moment we include them in the redux store. If we wouldn't have it in redux, this would be easier to understand I think)

However, when we fetch a single entry, e.g. by entering the page localhost/lab/substance/1, we want to add it to the byIds, so it doesn't need to be looked up again (at least for a while), but we don't want it to pop up in the visible list, because that all depends on if it should be in the current page of data. So if the list has populated samples in alphabetical order [sample1, sample2, sample3] and we suddenly get sample1000, we don't want it to show up in the list. That's up to other actions to decide.

GitEdvard: Ohh, that was quite an elaborate answer! You are probably right. Fetching a single entity is typically not for showing in a list, rather showing a detailed page for that entity.

withrocks: Yeah, I got carried away there :D

PR #169, opened 2020-08-13
Process definition

withrocks:


withrocks:
Draft until the shared stuff has been merged

src/sentry/static/sentry/app/redux/actions/processDefinition.js
```diff
@@ -1,350 +1,10 @@
...
-            disabled: false,
-            help: 'The method used for preparing the sample',
-            choices: ['Ready-made libraries', 'In-house libraries'],
-            required: true,
-          },
-          {
-            name: 'sample_type',
-            type: 'select',
-            label: 'Sample type',
-            choices: ['DNA', 'RNA', 'mRNA'],
```
GitEdvard, line 50: It seems to me that the demo stuff is now erased. Is that the intent?

GitEdvard (approved):
One thought came to my mind here. Consider these scenarios:
1. User choose from presets from a drop down box (preset = workflow = process?)
2. User view a single preset and dig into the particular task definitions within it
3. User dig in further and view all task instances (workbatch?) for a task definition
4. User views all samples for a single task instance/workbatch
...

Here it seems that different level of information is required to be fetched, depending on user case. Would that result in a multitude of action/reducer types, and would it still be possible to use the generalized action/reducers for this? On the other hand, I also think it would be premature to give to much concern for this at this stage. Just a thought.

withrocks:
Right! Yes, it's possible to use the generalized ones, because they are just similar to base classes (but it's all functions, which makes it not as familiar from an object oriented standpoint).

The idea is just to use the same code (for the most part) for all of those action/reducers, but they can then be customized.

You can think of it like that I didn't really remove any of the action/reducer code, in the sense that it's now all dynamically created on runtime rather than statically declared.

When I call:

```
export const processDefinitionActions = makeResourceActions(
  RESOURCE_NAME,
  '/api/0/process-definitions/'
);
```

all of those actions will be created just like before, but there is almost no code for this. If there are special cases, we can override by just assigning like this:

```
processDefinitionActions.someExtraFunctionRequiredForSubstances = () => ...
```

I hope that this answers the question!

src/sentry/static/sentry/app/redux/actions/processDefinition.js
withrocks: Right, this is just statically declared mock data that is now served for real by the server.

PR #170, opened 2020-08-14
Process assignment

withrocks:
Notice that I don't use the shared action/reducer here. That's just because it was a lot different and I didn't get to merging the two yet.

### DIFF
--- a/src/clims/api/endpoints/process_assignments.py
+++ b/src/clims/api/endpoints/process_assignments.py
@@ -16,11 +16,15 @@ class ProcessAssignmentsEndpoint(Endpoint):
         """
 
         # TODO-auth: Ensure that the user is only assigning samples that are under the organization
-        substances = request.data["substances"]
+
+        # Entities is a list of global ids (e.g. Substance-100)
+        entities = request.data["entities"]
         definition = request.data["definitionId"]
         variables = request.data["variables"]
 
         assignments = list()
-        assignments += self.app.workflows.batch_assign_substances(substances, definition, request.user, variables)
+
+        assignments += self.app.workflows.batch_assign(
+            entities, definition, request.user, variables)
 
         return Response({"assignments": len(assignments)}, status=201)
--- a/src/sentry/static/sentry/app/redux/actions/processAssignment.js
+++ b/src/sentry/static/sentry/app/redux/actions/processAssignment.js
@@ -1,0 +1,78 @@
+import {t} from 'app/locale';
+import {Client} from 'app/api';
+import {substanceSearchEntriesToggleSelectAll} from 'app/redux/actions/substanceSearchEntry';
+
+import {makeActionCreator} from 'app/redux/actions/shared';
+
+// Assigning substances to processes.
+// TODO-nomerge: Rename actions to processAssignments
+// TODO-nomerge: Use the wizard
+export const PROCESS_ASSIGN_SELECT_PRESET = 'PROCESS_ASSIGN_SELECT_PRESET';
+export const processAssignSelectPreset = makeActionCreator(
+  PROCESS_ASSIGN_SELECT_PRESET,
+  'preset'
+);
+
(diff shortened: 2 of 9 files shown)


src/clims/api/endpoints/process_assignments.py
```diff
@@ -16,11 +16,15 @@ def post(self, request, organization_slug):
         """
 
         # TODO-auth: Ensure that the user is only assigning samples that are under the organization
-        substances = request.data["substances"]
+
+        # Entities is a list of global ids (e.g. Substance-100)
+        entities = request.data["entities"]
```
GitEdvard, line 21: On a second thought, I think I like using "substances" more, rather than "entities". "Substances" communicate to a higher degree what a process is. Unpacking containers to get its contents seem to me like another concern that may be put somewhere else.

src/sentry/static/sentry/app/redux/actions/processAssignment.js
```diff
@@ -0,0 +1,78 @@
...
+// Assigning substances to processes.
+// TODO-nomerge: Rename actions to processAssignments
+// TODO-nomerge: Use the wizard
+export const PROCESS_ASSIGN_SELECT_PRESET = 'PROCESS_ASSIGN_SELECT_PRESET';
+export const processAssignSelectPreset = makeActionCreator(
+  PROCESS_ASSIGN_SELECT_PRESET,
+  'preset'
+);
+
+export const PROCESS_ASSIGN_SELECT_PROCESS_DEF = 'PROCESS_ASSIGN_SELECT_PROCESS_DEF';
```
GitEdvard, line 16: This action type is not used in an action

src/sentry/static/sentry/app/redux/actions/processAssignment.js
```diff
@@ -0,0 +1,78 @@
...
+export const processAssignSelectProcess = makeActionCreator(
+  PROCESS_ASSIGN_SELECT_PRESET,
```
GitEdvard, line 18: I think this row is wrong.

src/sentry/static/sentry/app/redux/reducers/processAssignment.js
```diff
@@ -0,0 +1,122 @@
...
+  assignPresetId: null,
+  assignProcessDefinitionId: null,
+  assignVariables: {},
+
+  editing: false,
+};
+
+const processAssignment = (state = initialState, action) => {
+  switch (action.type) {
+    case PROCESS_ASSIGN_SELECT_PRESET:
```
GitEdvard, line 29: In future, perhaps we could have a discussion whether this kind of state changes is necessary to have in action/reducers? It is a very simple state change, without api calls. Is the currently selected preset used in any other form outside of the preset form?

GitEdvard (approved):

src/clims/api/endpoints/process_assignments.py
withrocks: Right. I have a feeling that if we allow users to assign whatever (not only substances) it makes sense to name it that way here. It would perhaps be confusing if we start supporting assigning containers to then call it substances.

src/sentry/static/sentry/app/redux/reducers/processAssignment.js
withrocks: It's a good question. Actually the state of this preset stuff wasn't really that simple and I really liked using redux for it. I then simplified it because of a bug! But when it was complex, redux really shone.

In a nutshell it had to do with the different state that changes - if you select a preset, then the process and variables change, if you selected the process and variables, the preset changed (which was actually something I simplified later) and there were some corner cases there, which with a state framework like redux, I gotta say got very elegant.

The question about whether something should reside in redux or not is perhaps a hard one, I feel that it might be difficult to set clear rules about it. Let's discuss it later!

src/sentry/static/sentry/app/redux/actions/processAssignment.js
withrocks: Yes, I think there is some bug there. I hate to say it, but I think I'll fix it in a new PR.

PR #171, opened 2020-08-14
AssignToWorkflow view

withrocks:


src/sentry/static/sentry/app/components/substances/assignToWorkflow.jsx
```diff
@@ -3,30 +3,36 @@ import React from 'react';
...
 import {
   processAssignSelectPreset,
   processAssignSelectProcess,
   processAssignSetVariable,
-} from 'app/redux/actions/processAssign';
+  processAssignmentsPost,
+  processAssignmentsSetEditing,
+} from 'app/redux/actions/processAssignment';
+
+import {processDefinitionActions} from 'app/redux/actions/processDefinition';
```
GitEdvard, line 27: Throughout this file, "process" is used. But the name of the file is "assignToWorkflow".

src/sentry/static/sentry/app/components/substances/assignToWorkflow.jsx
```diff
@@ -139,23 +152,31 @@ const AssignToWorkflowButton = createReactClass({
...
+    });
+    const presetsSorted = sortBy(presets, 'label');
+
+    // TODO: Here we should handle the case where two workflows have the same name (but different
+    // namespaces), by showing the full name in that case
+    const processDefinitions = Object.entries(this.props.processDefinition.byIds).map(
       (entry) => {
-        return {value: entry[0], label: entry[1].name};
+        const processId = entry[1].id;
+        const elements = processId.split('.');
```
GitEdvard, line 169: I just want to check, should the process name be parsed out from process-id, and not process full name?

GitEdvard (approved):

src/sentry/static/sentry/app/components/substances/assignToWorkflow.jsx
withrocks: Yes :( I'm sorry. I hope that it will be easier to decide on all of these names after I've gotten the diagrams down!

src/sentry/static/sentry/app/components/substances/assignToWorkflow.jsx
withrocks: Ahh, right. Good catch. It's correct in this case as the ID here contains the name if I'm not mistaken. But it looks funky and just getting it from the name would be better :) A thing for the refactoring process!

PR #172, opened 2020-08-14
WorkBatch action reducers

withrocks:


src/sentry/static/sentry/app/redux/actions/workBatch.js
```diff
@@ -1,49 +1,11 @@
-import axios from 'axios';
+import {Client} from 'app/api';
```
GitEdvard: There are going to be a bunch of action reducers in future. What about organizing them, so that they are placed in the same folder as the main view calling them? This would be for a practical point of view. When updating them, I have otherwise have to look in 3 separate places which each are to be filled up with a lot of files.

src/sentry/static/sentry/app/views/workBatchDetails/shared/workBatchDetails.jsx
```diff
@@ -188,31 +51,22 @@ const WorkBatchDetails = createReactClass({
...
-      switch (this.state.errorType) {
-        case ERROR_TYPES.GROUP_NOT_FOUND:
-          return (
-            <div className="alert alert-block">
-              {t('The issue you were looking for was not found.')}
-            </div>
-          );
-        default:
-          return <LoadingError onRetry={this.remountComponent} />;
+  activeTab() {
```
GitEdvard, line 56: activeTab is not used.

src/sentry/static/sentry/app/views/workBatchList/index.jsx
```diff
@@ -1,48 +1,81 @@
...
-// import {Client} from 'app/api';
-// import {fetchOrgMembers} from 'app/actionCreators/members';
+import {workBatchActions} from 'app/redux/actions/workBatch';
+import ClimsTypes from 'app/climsTypes';
+import ListViewContainer from 'app/components/listViewContainer';
+import moment from 'moment';
+import Link from 'app/components/link';
 
-class WorkBatchList extends React.Component {
+class WorkBatchListContainer extends React.Component {
```
GitEdvard, line 10: The name of the file is different. Should it match?

GitEdvard (approved):

src/sentry/static/sentry/app/redux/actions/workBatch.js
withrocks: I wouldn't be against doing that, organizing both the actions and the views together based on main use case. This was setup by Mariya which I think prefers this, and I'm not against that either. So perhaps we can take it up again later.

src/sentry/static/sentry/app/views/workBatchDetails/shared/workBatchDetails.jsx
withrocks: Right! WorkBatchDetails is just there as a mock. I'll clean it up in my current task.

src/sentry/static/sentry/app/views/workBatchList/index.jsx
withrocks: I agree. I think it would be better if we synch it like you suggest, but currently this convention is in place in a lot of the Sentry code, i.e. the "connecting" component, the one that sets up things as required by react, is in a file called "index.jsx". I would not be against refactoring that all over, as I'm tired of figuring out where I am based on the path of the file rather than the name of the file!

PR #173, opened 2020-08-14
Fixes for TaskDefinition

withrocks:


src/clims/api/serializers/models/task_definition.py
```diff
@@ -0,0 +1,12 @@
...
+
+from rest_framework import serializers
+
+
+class TaskDefinitionSerializer(serializers.Serializer):
+    id = serializers.CharField(read_only=True)
+    name = serializers.CharField()
+    processDefinitionName = serializers.CharField(source="process_definition_name")
+    processDefinitionKey = serializers.CharField(source="process_definition_key")
+    taskDefinitionKey = serializers.CharField(source="task_definition_key")
```
GitEdvard, line 11: Perhaps just "key"?

src/clims/api/serializers/models/task_definition.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import absolute_import
+
+from rest_framework import serializers
+
+
+class TaskDefinitionSerializer(serializers.Serializer):
+    id = serializers.CharField(read_only=True)
+    name = serializers.CharField()
+    processDefinitionName = serializers.CharField(source="process_definition_name")
+    processDefinitionKey = serializers.CharField(source="process_definition_key")
```
GitEdvard, line 10: Perhaps use a navigation property instead? So using it would be mytask.processDefinition.name and mytask.processDefinition.key.

GitEdvard (approved):

src/clims/api/serializers/models/task_definition.py
withrocks: Good point! Since I really want to get this in, I'll just note it down and change it later.

src/clims/api/serializers/models/task_definition.py
withrocks: Yes, that sounds better! I'll add this to a cleanup PR.

PR #174, opened 2020-08-14
Tasks

withrocks:


src/sentry/static/sentry/app/views/tasks/index.jsx
```diff
@@ -0,0 +1,108 @@
...
+    super(props);
+  }
+
+  componentDidMount() {
+    const {processKey, taskKey} = this.props.routeParams;
+
+    // If we don't have the task definition, we load that too:
+    // TODO: check if we have it first
+    this.props.getTaskDefinition(processKey, taskKey);
+    this.props.getTasks(this.props.organization, processKey, taskKey);
```
GitEdvard, line 22: The parameter "taskKey" indicates that there should be only a single instance returned from this this method? But the method name "getTasks" says that there are many.

GitEdvard (approved):

src/sentry/static/sentry/app/views/tasks/index.jsx
withrocks: Right, they are many, but if you send in the search parameters processKey and taskKey, you'll always get a single instance. We might want to treat them as a key everywhere, i.e. if `processKey` is "process" and `taskKey` is "taskKey" then perhaps we should use "processKey/taskKey" as the full key for the task. But for now it copies how Camunda exposes it.

withrocks:
I think it's ok to merge this one!

PR #182, opened 2020-09-07
Prevent multiple entries of identical Location in db

GitEdvard:
Purpose:

As the header says. During script execution, save() may be called multiple times for a single sample, whether or not the sample has been updated. This bugfix prevents additional identical entry to be created for each save() call.

### DIFF
--- a/src/clims/services/extensible.py
+++ b/src/clims/services/extensible.py
@@ -135,6 +135,46 @@ class ExtensibleService(object):
         return extensible_type
 
 
+class Location(object):
+    """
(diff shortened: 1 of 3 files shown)


src/clims/services/extensible.py
```diff
@@ -135,6 +135,46 @@ def _register_model(self, name, org, plugin, property_types=None):
         return extensible_type
 
 
+class Location(object):
```
withrocks, line 138: Looks good, but I wonder if we can merge it with other code that exists. But that's for another commit!

withrocks (approved):

PR #183, opened 2020-09-08
clims-407, add handling for project in substance.create_child()

GitEdvard:
By default, create_child() inherits the parent's project. Project may also be overridden.


withrocks (approved):
Cool!

PR #186, opened 2020-09-22
Wait for user tasks via Workflow def

withrocks:
Adds a helper method for waiting for user tasks in Camunda to reach a certain level.

### DIFF
--- a/src/clims/services/workflow/__init__.py
+++ b/src/clims/services/workflow/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 
+from retry import retry
 import logging
 import six
 from clims import utils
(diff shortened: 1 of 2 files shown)


src/clims/services/workflow/__init__.py
```diff
@@ -291,6 +308,7 @@ def get_tasks_by_ids(self, task_ids):
         task_ids_by_provider = defaultdict(list)
 
         for task_id in task_ids:
+            print(task_id, "HERE")
```
GitEdvard, line 311: This seems to be a debug line. Should it be removed?

GitEdvard (approved):

src/clims/services/workflow/__init__.py
withrocks: Yes :| thanks, i'll fix that.

PR #187, opened 2020-09-22
Fully qualify "formKeys" from Camunda

withrocks:


src/clims/services/workflow/camunda/client.py
```diff
@@ -159,22 +159,35 @@ def unsafe_delete_all_deployments(self):
...
+        it will always have a fully qualified name so they will not clash with other plugins.
+
+        Example: User creates a diagram with the ID "SequenceSimple" and then registers it
+        through the plugin clims.plugins.demo.dnaseq, the process will actually be registered
+        as clims.plugins.demo.dnaseq.SequenceSimple. Furthermore, all calls to to subprocesses
+        will be made to fully qualified names using the same module.
+
+        Form keys are also changed to the fully qualified version.
+
+        If the user does fully qualify names (they contain a dot), no change takes place.
```
GitEdvard, line 172: "ID" and "formKey" are here used interchangeable, although it's only "formKey" that are handled in this method. I'm not certain of what the purpose of formKey is. It seems to be an identifier of the workflow. But what about the "form" in the formKey?

src/clims/services/workflow/camunda/client.py
```diff
@@ -204,6 +217,14 @@ def is_relative(name):
                 call_activity.attrib[
                     "calledElement"] = called_process_id_fully_qualified
 
+        # Qualify all form keys
```
GitEdvard, line 220: According to my taste, I prefer much shorter methods than this. I had to scroll to find out where "namespace" comes from (input argument).

src/clims/services/workflow/camunda/client.py
```diff
@@ -204,6 +217,14 @@ def is_relative(name):
...
+        user_tasks = process.findall("bpmn:userTask", ns)
+        for user_task in user_tasks:
+            form_key = user_task.attrib.get("{http://camunda.org/schema/1.0/bpmn}formKey", None)
+            if form_key and is_relative(form_key):
+                form_key_qualified = "{}.{}".format(namespace, form_key)
+                user_task.attrib["{http://camunda.org/schema/1.0/bpmn}formKey"] = form_key_qualified
```
GitEdvard, line 226: It took me a while before I realized that this method mutates the input argument. It's probably necessary, athough this kind of mutation is nasty. It would be nice if this side effect was expressed or made clear in some way.

GitEdvard:
I have some minor comments.

GitEdvard (approved):

src/clims/services/workflow/camunda/client.py
withrocks: A formkey is a Camunda feature that has the id of the form (view) that should be used in the UserTask (https://docs.camunda.org/manual/7.8/user-guide/task-forms/). We need some way to connect the UserTask and the view we're showing. We could also use the UserTask ID, but that would mean that we couldn't reuse views for different user tasks, because those IDs need to be unique.

src/clims/services/workflow/camunda/client.py
withrocks: I agree! I can refactor it when I have time. It's often more clear when reading it than when writing it.

src/clims/services/workflow/camunda/client.py
withrocks: Ahh. The name of the method is `refine_xml` so I would say that it indicates that it's modifying it. Tell me if you think it requires more than that though!

src/clims/services/workflow/camunda/client.py
GitEdvard: :+1:

src/clims/services/workflow/camunda/client.py
GitEdvard: Yes, I know that the name indicates it, but I think in general that modifying an input argument is the least obvious way to mutate data. I'm not strongly against this method above, but I still wanted to point out my opinion about it :)

withrocks: Right, I get your point and I agree with it totally.

PR #191, opened 2020-09-22
Add a test helper method for fetching api resp. of an exp. length

withrocks:


src/sentry/testutils/cases.py
```diff
@@ -496,6 +497,19 @@ def get_url(self, org=True):
             kwargs["organization_slug"] = self.organization.slug
         return reverse(self.endpoint, kwargs=kwargs)
 
+    def wait_for_endpoint_list(self, url, num, tries=3, delay_sec=0.5):
```
GitEdvard: When using this method, it would be much more expressive to have "expected_num" instead of "num"

GitEdvard (approved):

src/sentry/testutils/cases.py
withrocks: Good point!

PR #192, opened 2020-09-22
clims-419, Workbatchfile available from workbatch by file handle

GitEdvard:
Purpose:

As the header says. In this case, it's going to be used in the "Analyze Quality Table" in snpseq. Here, the scripts want access to a file holding quality data, regardless of its name. To achieve this, the file is labeled with a file-handle at creation. 

Implementation:
I created classes for WorkBatch and WorkBatchFile in the service layer, that in turn uses the corresponding django model classes. By doing this, I can import other references from the service layer, without having the "python import short-curcuit" problem, that happens when importing service classes to the django model classes.


src/clims/services/workbatch.py
```diff
@@ -12,3 +15,73 @@ def register_step_template(self, step_template_cls):
     def get_step_template(self, name):
         from clims.utils import single
         return single([step for step in self.step_templates if step.name == name])
+
+
+class WorkBatch(object):
+    def __init__(self, work_batch_model):
+        self.archetype = work_batch_model
```
withrocks, line 22: Looks cool all of this.

I think we might consider making this into an Extensible, as I think you've brought up. That would make sense since it allows us to define step level properties.

I'm imagining a class hierarchy like this:
-> WorkBatch(ExtensibleBase)
--> FragmentAnalyzeWork(WorkBatch)

But I think that can follow later!

GitEdvard: Yes, I have also been into that idea.

withrocks (approved):

PR #195, opened 2020-09-24
Add WorkUnit model and add containers to WorkBatches

withrocks:


src/clims/models/workunit.py
```diff
@@ -0,0 +1,80 @@
...
+from django.db import models
+from sentry.db.models import (FlexibleForeignKey, sane_repr)
+from clims.models.substance import Substance
+from clims.models.container import Container
+
+
+class WorkUnit(models.Model):
+    """
+    Represents a single item that needs to be processed manually. A WorkBatch consists of 1..n
+    WorkUnits.
```
GitEdvard, line 12: I'm not sure what you mean with "processed manually". I guess you refer to the real world here?

src/clims/models/workunit.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # The WorkType is the full namespace of the Python class that defines what happens in the
+    # workflow
+    work_type = models.TextField(null=True)
+
+    @property
+    def flexible_id(self):
+        """
+        Because a WorkUnit can either represent an external (only in workflow engine) or internal
+        (also saved locally), when creating, the user sends in the "token". The user should
```
GitEdvard: This sentence is incomplete and is a bit hard to understand. "represent an external ..." (external what?)

src/clims/models/container.py
```diff
@@ -17,13 +17,15 @@ class Container(ExtensibleModel):
...
     # TODO: Benefits of FlexibleForeignKey
-    # TODO: Decide on naming for related_name
     container_type = FlexibleForeignKey('ContainerType', null=True, related_name="containers")
 
     name = models.TextField(null=True)
 
     organization = FlexibleForeignKey('sentry.Organization')
 
+    # The work batch in which this container was created, if it was created in a work batch context.
+    work_batch = FlexibleForeignKey("WorkBatch", null=True, related_name="output_containers")
```
GitEdvard, line 27: The association between container and workbatch is defined in two ways, 1) from container to workbatch, 2) from container to workunit to workbatch. In the end, I think this might lead to confusing code. Imagine that assiciation 2) is used almost all the time, except in some corner cases where association 1) is used. I can smell bugs coming here.

src/clims/models/workunit.py
```diff
@@ -0,0 +1,80 @@
...
+
+class WorkUnit(models.Model):
+    """
+    Represents a single item that needs to be processed manually. A WorkBatch consists of 1..n
+    WorkUnits.
+
+    WorkUnits can either point to an external workflow engine (such as Camunda, where they map
+    to UserTasks) or not. If they point to no workflow engine, resolving the WorkUnit will just
+    close that item without further tracking. This makes it possible to use the same views to
+    work on items that are within a workflow or not.
```
GitEdvard, line 17: This is hard to understand without examples.

src/clims/models/workunit.py
```diff
@@ -0,0 +1,80 @@
...
+    work_batch = FlexibleForeignKey('clims.WorkBatch', null=True, related_name="work_units")
+
+    # A WorkUnit either connects to a single Substance or a Container, never both.
+    substance = FlexibleForeignKey('clims.Substance', null=True)
+
+    container = FlexibleForeignKey('clims.Container', null=True)
+
+    # The WorkType is the full namespace of the Python class that defines what happens in the
+    # workflow
+    work_type = models.TextField(null=True)
```
GitEdvard, line 40: This would be nice to see in a test. It would be no question of what work_type is and how it's defined.

GitEdvard (commented):
It would be much easier to give meaningful comments if I could see this code used in a context (read tests). workunit contains a bunch of different ids that I dont really can relate to. (external_work_unit_id, flexible_id, tracked_object_global_id). How are they used within clims?

src/clims/models/workunit.py
withrocks: That means that a human will have to work on it, rather than something that happens automatically within a workflow process.

src/clims/models/workunit.py
withrocks: Right! I thought I had stopped writing this kind of comments, sorry for not catching it before you saw it :) I'll fix it

src/clims/models/container.py
withrocks: Right. This is just a pointer to the containers that were created within the workbatch. We'll need them for the UI to be able to show which containers are shown on the right in the Transition component.

The containers in WorkUnits are the input containers, but only in the case if someone is writing a workflow that tracks containers rather than substances.

I think that the dangers you're talking about are difficult to guard against more while data modelling. If the link is required we'll have to have it. But we can use naming to help, and I think we are doing that because the link in the work_batch is called `output_containers`. We can try to guard against wrong usage more by having good documentation and a clear API in the service layer.

src/clims/models/workunit.py
withrocks: Right. It's just the full name of a Python class, the one that is currently called `Step` but will be called `Work`. There will be tests for this for sure (but in the service layer).

src/clims/models/workunit.py
withrocks: Yes, it's a catch-22. If I send in all the context it might be too much info at once :). But it'll come in the ~next PR

withrocks:
About those ids. I removed the `flexible_id` as it's not required on the model. It comes into play in a later PR.

The other IDs you mention are for different objects:
 * external_work_unit_id: The id of the WorkUnit in the workflow engine (Camunda for now)
 * tracked_object_global_id: The id of whatever object it is that we wish to process. Usually this will be a substance and they will have a "global_id" on the form "Substance-<unique number>".

src/clims/models/workunit.py
GitEdvard: ok

src/clims/models/container.py
GitEdvard: Now I see that this association is for the output containers, and the other is for the input containers, and that changes the situation a bit. Probably Camunda is only tracking input samples/containers, right? I still think, however, that this violates the "principle of least surprises".

src/clims/models/workunit.py
GitEdvard: ok, that was what we talked about in todays standup :)

src/clims/models/workunit.py
GitEdvard: ok

PR #196, opened 2020-09-28
Remove clarity-ext legacy directory

GitEdvard:
Purpose:

Since we are importing clarity-ext references directly from the real clarity-ext repository, I don't see any purpose with having this legacy directory. This directory also have hardened my work somewhat, since all clarity-ext file exists in doublets, and I sometimes navigate to the wrong clarity-ext file during development.


withrocks:
Yes excellent, this was added for reference as a total hack and some point and was not supposed to remain.

PR #198, opened 2020-09-29
Make sure that clean removes the current config files

withrocks:
Running `make fresh` will now make sure that the user has new
config files installed.


mariya (approved):
Looks good!

PR #202, opened 2020-10-12
Send POST request with action creators

GitEdvard:
Purpose:
Include action creator POST requests in tests. 

Implementation:
Call POST by axios.post() so that it may be mocked in tests.


mariya (approved):
Looks great - good job! This is going to be really useful.

GitEdvard:
Thanks for looking!

PR #203, opened 2020-10-13
CLIMS-432 - adjust headers depending on what the current grouping is

mariya:
This still feels a little too hardcoded to me, but it gets the job done.

### DIFF
--- a/src/sentry/static/sentry/app/views/substances/substances.jsx
+++ b/src/sentry/static/sentry/app/views/substances/substances.jsx
@@ -74,57 +74,89 @@ class Substances extends React.Component {
   }
 
   getHeaders() {
-    // TODO: This should be returned as a contract from the plugin that is registered for this.
-    return [
-      {
-        Header: 'Sample name',
-        id: 'name',
-        accessor: (d) => d.name,
-        fontstyle: (isGroupHeader) => (isGroupHeader ? 'italic' : 'normal'),
-      },
-      {
-        Header: 'Container',
-        id: 'container',
-        accessor: (d, isGroupHeader) =>
-          isGroupHeader ? null : d.location ? d.location.container.name : '<No location>',
-      },
-      {
-        Header: 'Index',
-        id: 'index',
-        accessor: (d, isGroupHeader) =>
-          isGroupHeader ? null : d.location ? d.location.index : '<No location>',
-      },
-      {
-        Header: 'Volume',
-        id: 'volume',
-        accessor: (d, isGroupHeader) =>
-          isGroupHeader
-            ? null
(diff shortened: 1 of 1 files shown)


src/sentry/static/sentry/app/views/substances/substances.jsx
```diff
@@ -74,57 +74,89 @@ class Substances extends React.Component {
...
+        nameHeaderName = 'Container';
+        break;
+      case 'sample_type':
+        nameHeaderName = 'Sample Type';
+        break;
+      default:
+        nameHeaderName = 'Name';
+    }
+
+    let nameHeader = {
```
GitEdvard, line 92: Yes, it's a little more pleasant to read when split up in variables.

src/sentry/static/sentry/app/views/substances/substances.jsx
```diff
@@ -74,57 +74,89 @@ class Substances extends React.Component {
...
-        id: 'days_waiting',
-        accessor: (d, isGroupHeader) => (isGroupHeader ? null : d.days_waiting),
-      },
-    ];
+    // TODO: Headers should be returned as a contract from the plugin that is registered for this.
+    let {groupBy} = this.props.substanceSearchEntry;
+
+    let nameHeaderName = '';
+    switch (groupBy) {
+      case 'container':
```
GitEdvard, line 82: Here I wonder if you have considered that *samples* are shown even for group by = container, when user clicking on the expand/collapse tick beside the container row? If you have a plan for this, I think this is ok (or if you have missed it, you probably have to consider it). By the way, I myself have not figured out how to present this expandable list view. It would be very appreciated if you could come up with something. Maybe you would like to discuss this?

GitEdvard (commented):
Looks good! I have a comment. Please tell me when you want to merge.

src/sentry/static/sentry/app/views/substances/substances.jsx
mariya: Yeah I gave it some thought. I think like this: when the user lands on the default view, they see a list of ungrouped samples and the first header reads "Sample name". Then, when they subsequently select a Container/Sample Type grouping it is pretty clear from a UX perspective that what you see is groups of samples. So I think adding an additional "Sample Name" header will only clutter things and isn't necessary from an informational standpoint. Maybe @withrocks has thoughts on this as well? If you're ok with this, let's merge please!

GitEdvard: Yes, I see that you already have considered this. So this is OK with me! I'll merge.

PR #204, opened 2020-10-14
447 show buttons in workbatch according to configuration

GitEdvard:
Purpose:
Show buttons in a workbatch view according to a configuration provided by the api. 

Scope:
This is an implementation in the UI only. The corresponding endpoints /work-configurations and /events is not yet implemented in the backend. Furthermore, since a workbatch page is not yet in place, I started with one here and called it "example workbatch". This page only handles buttons, nothing else. 

Implementation:
I use the general actions creators in shared.js to create actions for the button clicked events as well as fetching a work configuration. 

Note:
I left some method name changes in workBatchDetials from previous changes. Most of them were changed back. Please have a look if these method names were the intended. They were inconsistent when I found them ("get" in the beginning or in the middle of the method name).

### DIFF
--- a/src/sentry/static/sentry/app/redux/actions/shared.js
+++ b/src/sentry/static/sentry/app/redux/actions/shared.js
@@ -95,10 +95,10 @@ const acGetFailure = (resource) =>
   makeActionCreator(`GET_${resource}_FAILURE`, 'statusCode', 'message');
 
 const acGet = (resource, urlTemplate) => {
-  return (id) => (dispatch) => {
+  return (org, id) => (dispatch) => {
     dispatch(acGetRequest(resource)(id));
-
-    const url = urlTemplate.replace('{id}', id);
+    let url = urlTemplate.replace('{org}', org.slug);
+    url = url.replace('{id}', id + '/');
     return axios
       .get(url)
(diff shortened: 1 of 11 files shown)


mariya (changes requested):
The code looks fine, but I have some questions/comments about naming. Is the intention that ExampleWorkbatchContainer will at some point be edited to include the real implementation? If so, I think it is better to just name it WorkbatchContainer directly and just add TODO comments where applicable. If the plan is to replace ExampleWorkbatchContainer entirely in the future, then perhaps DummyWorkbatchContainer is a clearer name for something intended to be thrown away?

GitEdvard:
Yes, my intention was that ExampleWorkbatch would be used in future. So I'll rename it. Thanks for the comment.

GitEdvard:
I discovered that there are already an existing workbatch details container, and unfortunately this cannot be shown as the current state of the clims. Therefore I renamed my view to workBatchDetailsWaitingToBeMerged. (It will now show up next to the original view which is good). I also split up the code into a Container and Presentational class, in a similar manner as other views.

mariya (approved):
This looks good to me, feel free to merge!

PR #205, opened 2020-10-20
328 workbatch details

withrocks:


src/clims/api/endpoints/available_work.py
```diff
@@ -2,16 +2,18 @@
 
 from sentry.api.paginator import OffsetPaginator
 from sentry.api.bases.organization import OrganizationEndpoint
-from clims.api.serializers.models.task_definition import TaskDefinitionSerializer
+from clims.api.serializers.models.work_definition import WorkDefinitionSerializer
 
 
-class TaskDefinitionEndpoint(OrganizationEndpoint):
+class AvailableWorkEndpoint(OrganizationEndpoint):
```
withrocks, line 8: Change to `AvailableWorkDefinitionsEndpoint`

src/clims/api/endpoints/available_work.py
```diff
@@ -2,16 +2,18 @@
...
+    name = 'clims-api-0-available-work'
+
     def get(self, request, organization):
-        task_definitions = self.app.workflows.get_task_definitions()
+        work_definitions = self.app.workflows.get_work_definitions()
```
withrocks, line 12: Why not "available" work definitions?

src/clims/api/endpoints/work_batch.py
```diff
@@ -29,7 +29,7 @@ def post(self, request, organization):
         logger.debug("Creating workbatches for task list: {}".format(tasks))
 
         # Creating a workbatch takes a list of tasks that are ready for a work batch:
-        work_batch = self.app.workflows.create_work_batch(tasks, organization)
+        work_batch = self.app.workflows.start_work(tasks, organization)
```
withrocks, line 32: Start_work_batch.

src/clims/api/endpoints/work_definition_details.py
```diff
@@ -0,0 +1,24 @@
...
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+
+from sentry.api.base import Endpoint, DEFAULT_AUTHENTICATION
+from clims.utils import single_or_default
+from clims.api.serializers.models.work_definition import WorkDefinitionSerializer
+from clims.api.serializers.models.workunit import WorkUnitSerializer
+
+
+class WorkDefinitionDetailsEndpoint(Endpoint):
```
withrocks, line 12: Which is used?!

src/clims/api/endpoints/work_units.py
```diff
@@ -0,0 +1,49 @@
...
+from sentry.api.bases.organization import OrganizationEndpoint
+
+from sentry.api.base import DEFAULT_AUTHENTICATION, Endpoint
+
+from clims.api.serializers.models.workunit import WorkUnitSerializer
+
+logger = logging.getLogger(__name__)
+
+
+class WorkUnitsEndpoint(OrganizationEndpoint):
```
withrocks, line 16: Add documentation that says what WorkUnit really is (instance of work on one "tracked object", usually Substance)

src/clims/api/endpoints/available_work.py
```diff
@@ -2,16 +2,18 @@
 
 from sentry.api.paginator import OffsetPaginator
 from sentry.api.bases.organization import OrganizationEndpoint
-from clims.api.serializers.models.task_definition import TaskDefinitionSerializer
+from clims.api.serializers.models.work_definition import WorkDefinitionSerializer
```
GitEdvard, line 5: Here, I think there is an unfortunate name collision between work_definition and work_configuration. "definition" and "configuration" are very similar words but they mean different things here in clims. work_configuration is more of a UI thing, what's should be showed when entering a workbatch. "work_definition", is so far used in the use case to show available work, ie. work with queued sample into it, which has not yet been started. In this specific use case, we just need the name of the type of work to be done ("fragment analyze").

src/clims/api/endpoints/work_units.py
```diff
@@ -0,0 +1,49 @@
...
+        work_definition_key = request.GET.get("workDefinitionKey")
+
+        work_units = self.app.workflows.get_work_units(work_definition_key,
+                                             process_definition_key)
+        print("stuffo", work_units)
+        ret = WorkUnitSerializer(work_units, many=True).data
+        return Response(ret)
+
+
+class WorkUnitsByWorkDefinitionEndpoint(Endpoint):
```
GitEdvard, line 33: This name is precise, but it's also a bit long and cumbersome to read. In the end, I think I would like to see a more striking word than "work definition", since I probably have to think a while before I get the meaning of it. I think a good measure is to use it in a sentence for an use case... "User clicks on 'available work' and sees a list of (???), each (???) is showed with the number of samples that are queued for it". Here I would like a short, striking and not necessary a technical word that a lab user could use spontaneously. Maybe task, user task, or something alike?

I don't think it adds value in itself to name each concept with the word "work" in it, even if it's related to available work. I would rather accentuate the difference between workbatch, wich always is an instance with a specific set of samples, a timestamp, user name on one hand, from a work definition on the other hand, which is just a kind of recipe for a workbatch.

As far as I understand, "work unit" only has meaning together with a work definition, or a workbatch. Otherwise one would use the endpoint for substances, or containers, etc directly, right? 

An alternative to this endpoint would be to only use work-units, with a paramater for work definition. I like the more verbose version like that above, however.

src/clims/api/endpoints/work_definition_details.py
GitEdvard: For me, WorkDefninitionDetail would be the unit that is used when user enters a new or existing workbatch, containing a list of substasks for one thing. On the other hand, the WorkDefinitions endpoint would be used for showing a list of work definitions. I think we already agree on that. I just want to make things clear, and to avoid misunderstandings :)

GitEdvard: Here, things get messy however. Exactly what fields should be included in work definition, workbatch and work configuration respectively? Take properties in work-configuration for example, (which is step udfs in clarity). Work configuration is at present thought of to be a field within a workbatch, so that all extensible properties may be lumped into one single field. In principle though, as far as my understanding goes, these fields could equally well be included in work definition? Maybe work-defnition and work-configuration is really the same concept?

GitEdvard: There is another thing that came into my mind here. I make comments about the endpoints, but I think it's a slight difference between the endpoints and the domain model. In my mind, endpoints is much more liable to adjust to the demands at the UI, while the domain models aims for solving domain logic (do we agree on that?).

GitEdvard (commented):
Here I wrote down some of the thoughts that came into my mind. Hopefully it will serve as a basis of a good discussion :)

src/clims/api/serializers/models/process_definition.py
```diff
@@ -15,7 +15,7 @@ class ExtensibleBaseFieldSerializer(serializers.Serializer):
     help = serializers.CharField()
     type = serializers.CharField()
     # TODO: It would be nice to change display_name to label everywhere
-    display_name = serializers.CharField()
+    label = serializers.CharField()
```
GitEdvard, line 18: Ok, an alternative way to do this is label  = serializers.CharField(source='display_name')

src/clims/services/workflow/__init__.py
```diff
@@ -279,45 +280,62 @@ def install(self, workflow_cls, plugin_reg):
...
         batch.save()
 
-        for task in tasks:
-            task.tracked_object._archetype.work_batches.add(batch)
-            task.tracked_object._archetype.save()
+        # 3. Add all the work units to the batch
+        for work_unit in work_units:
+            work_unit.work_batch = batch
+            work_unit.save()
+            print("ME HERE", work_unit, work_unit.tracked_object)
```
GitEdvard: NO, no prints! :)

src/clims/services/workflow/__init__.py
```diff
@@ -279,45 +280,62 @@ def install(self, workflow_cls, plugin_reg):
...
+        for work_unit in self._materialize_external_work_units(work_units):
+            print(">>>2<<<", id(work_unit), work_unit.id, work_unit.tracked_object, work_unit.substance)
+
         # 1. Fetch all the tasks from the task IDs we received
-        tasks = self.get_tasks_by_ids(task_ids)
-        name = list({t.name for t in tasks})
-        if len(name) > 1:
-            raise AssertionError("Expecting only one task name when creating a work batch, got: '{}'".format(name))
-        name = name[0]
+        work_types = list({t.work_type for t in work_units})
```
GitEdvard: work_types is a little misleading, since it's about work-unit-type. On the other side, it's cumbersome with long variable names. But this method is rather long, 40 lines (?), and when it's long I'm inclined to use more descriptive variable names.

src/clims/services/workflow/__init__.py
```diff
@@ -279,45 +280,62 @@ def install(self, workflow_cls, plugin_reg):
...
-            task.tracked_object._archetype.save()
+        # 3. Add all the work units to the batch
+        for work_unit in work_units:
+            work_unit.work_batch = batch
+            work_unit.save()
+            print("ME HERE", work_unit, work_unit.tracked_object)
+
+        # Fire the on_created event
+        work_type_instance = self._app.plugins.handlers.init_by_name(work_type, batch)
+        work_type_instance.on_created()
```
GitEdvard: In most cases this is "substance.on_created". It seem to me a little far reached. Is it possible to make this encapsulated in some way?

src/clims/services/workflow/__init__.py
```diff
@@ -391,44 +412,76 @@ def batch_get_tracked_objects(self, category, keys):
...
-        tasks_grouped_by_category = defaultdict(list)
-        tasks_grouped_by_global_id = defaultdict(list)
-        for task in tasks:
-            tasks_grouped_by_category[task.tracked_object_category].append(
-                task)
-            tasks_grouped_by_global_id[task.tracked_object_global_id].append(
-                task)
+    def _materialize_external_work_units(self, work_units):
+        """
+        Given a list of work_units (which can be external or regular), makes sure
```
GitEdvard: I don't know the difference between regular and external work unit, so this is a reminder that I should ask you about it!

src/clims/services/workflow/__init__.py
```diff
@@ -391,44 +412,76 @@ def batch_get_tracked_objects(self, category, keys):
...
+        that all ExternalExtensibleBase have materialized tracked object.
+
+        Ignores regular WorkUnits.
```
GitEdvard: I dont like that the input argument is mutated.

src/clims/services/workflow/__init__.py
```diff
@@ -391,44 +412,76 @@ def batch_get_tracked_objects(self, category, keys):
...
-                    task.tracked_object = tracked_object
-        return tasks
-
-    def get_tasks(self, task_definition_key=None, process_definition_key=None):
-        # TODO: If we ever have more than one workflow engine, fetching asynchronously would make
-        # a lot of sense here
+                for work_unit in external_work_units_grouped_by_global_id[tracked_object.global_id]:
+                    work_unit.tracked_object = tracked_object
+
+    def get_work_units(self, work_definition_key=None, process_definition_key=None):
```
GitEdvard, line 455: Do we have "process" as concept within clims? As I recall, in camunda a process is associated with a single work unit. If it's a camunda concept only, I think that it should be a small translation layer, rather than spread out in this rather large file.

src/clims/services/workflow/__init__.py
```diff
@@ -391,44 +412,76 @@ def batch_get_tracked_objects(self, category, keys):
...
+        self._materialize_external_work_units(ret)
+        print(">>>", [x.__dict__ for x in ret])
+        return ret
 
-        return all_tasks
+    def get_work_units_by_flexible_ids(self, flexible_ids):
+        """
+        Maps from a list of flexible WorkUnit, which are IDs that either point
+        to the workflow engine (if the WorkUnit doesn't exist locally yet) or
+        to a
```
GitEdvard: This sentence stops right into nowhere.

src/clims/services/workflow/__init__.py
```diff
@@ -391,44 +412,76 @@ def batch_get_tracked_objects(self, category, keys):
...
+        """
+        raise NotImplementedError()
+        flexible_ids_by_provider = defaultdict(list)
```
GitEdvard: I'm not fond of flexible ids, because one have to go deep into the details to understand what it is. Here I suspect that work units are to be fetched in two different scenarios, 1) samples that are in queue for a work definition, 2) samples that are included in a workbatch. Is it so? If that is the case, I would prefer to have two methods for this, and there would be no need to have flexible id.

src/clims/services/workflow/__init__.py
```diff
@@ -512,84 +565,94 @@ def batch_assign_containers(self, containers, process, user, variables):
...
         if process_definition_key:
-            task_definitions = task_definitions.filter(process_definition__key=process_definition_key)
-        if task_definition_key:
-            task_definitions = task_definitions.filter(task_definition_key=task_definition_key)
-
-        task_definitions_with_instance_count = task_definitions.values(
-            'task_definition_key', 'name', 'process_definition__name',
+            camunda_task_definitions = camunda_task_definitions.filter(process_definition__key=process_definition_key)
+        if work_definition_key:
+            camunda_task_definitions = camunda_task_definitions.filter(task_definition_key=work_definition_key)
```
GitEdvard, line 631: I would like that camunda lingo and clims lingo were only to be co-existing within a thin translation layer. This file is rather big, and the two different lingos are spread out in a large area.

src/clims/services/workflow/__init__.py
```diff
@@ -512,84 +565,94 @@ def batch_assign_containers(self, containers, process, user, variables):
...
         self.count = count
 
 
-class ProcessTask(object):
-    """Represents a single instance of a step running in a workflow process."""
+class WorkflowProcessNotFound(Exception):
+    pass
+
+
+class ExternalWorkUnit(object):
```
GitEdvard: I would like a short presentation of how this mechanism works!

src/clims/services/workflow/camunda/client.py
```diff
@@ -56,17 +57,17 @@ def start_workflow(self, workflow, item):
...
             raise UnexpectedHttpResponse(json["message"], response.status_code)
 
-    def get_tasks_by_ids(self, task_ids):
+    def get_work_units_by_ids(self, task_ids):
         # TODO-perf: We must be able to fetch this in a batch call. Look into setting up
         # https://github.com/camunda/camunda-bpm-graphql (preferably) or direct DB access to resolve
         # this, as it's expected that these will be many calls
         # TODO: paging
-        from clims.services.workflow import ProcessTask
+        raise NotImplementedError()
```
GitEdvard: I understand that you keep the code below because it's "good to have"? I also have this habit. But I have got the advise to get rid of not used code ("zombie code"), and if one would like to retrieve it again, it's should still be accessible from git. What do you think about that? I personally don't have much experience to find old code from git, and cannot tell if it's practically doable or not.

src/clims/services/workflow/__init__.py
```diff
@@ -391,44 +412,76 @@ def batch_get_tracked_objects(self, category, keys):
...
+        return ret
 
-        return all_tasks
+    def get_work_units_by_flexible_ids(self, flexible_ids):
+        """
+        Maps from a list of flexible WorkUnit, which are IDs that either point
+        to the workflow engine (if the WorkUnit doesn't exist locally yet) or
+        to a
+        """
+        raise NotImplementedError()
```
GitEdvard: Perhaps you should give a comment of whether this is a method that is to be implemented in future, or if it's an old method not to be used anymore?

tests/clims/api/endpoints/test_available_work.py
```diff
@@ -0,0 +1,57 @@
...
+        self.create_available_work(count)
+
+        self.login_as(self.user)
+
+        url = reverse(AvailableWorkEndpoint.name, args=(self.organization.name, ))
+        response = self.wait_for_endpoint_list(url, 1)
+
+        assert response.status_code == 200
+        assert response.json() == [{
+            u'id': u'clims.plugins.demo.dnaseq.workflows.sequence.SequenceSimple:DataEntry',
```
GitEdvard, line 51: I dont get the meaning of DataEntry

GitEdvard (commented):
(workflow service uses WorkUnits) I leave these comments, so that you may write tickets to fix things in retrospect, so that you can merge it now.

tests/clims/api/endpoints/test_task_definition.py
```diff
@@ -1,15 +1,15 @@
 from __future__ import absolute_import
 
 from sentry.testutils import APITestCase
-from clims.plugins.demo.dnaseq import DemoDnaSeqPlugin
+from clims.api.endpoints.work_unit_definition import WorkDefinitionEndpoint
 
 
-class TaskDefinitionTest(APITestCase):
-    endpoint = 'clims-api-0-task-definitions'
+class WorkDefinitionTest(APITestCase):
```
GitEdvard: The name of the file dont seem to match the class name, WorkDefinitionTest and test_task_definition

tests/clims/api/endpoints/test_available_work.py
```diff
@@ -0,0 +1,57 @@
...
+
+from sentry.testutils import APITestCase
+
+from clims.plugins.demo.dnaseq import DemoDnaSeqPlugin
+from clims.plugins.demo.dnaseq.models import ExamplePlate, ExampleSample
+from clims.plugins.demo.dnaseq.workflows.sequence import SequenceSimple
+from clims.api.endpoints.available_work import AvailableWorkEndpoint
+
+
+class AvailableWorkEndpointTest(APITestCase):
```
GitEdvard, line 15: There is a name collision between available work and word definitions. What is showed in the "Available work" in UI is a list of word definitions. So it seems wrong to have tests for both available work and word definitions.

tests/clims/api/endpoints/test_work_definition_details.py
```diff
@@ -0,0 +1,43 @@
...
+            sample = ExampleSample(name="sample-{}".format(x))
+            cont.append(sample)
+        cont.save()
+        workflow = SequenceSimple()
+        workflow.assign(cont)
+
+        # Get through REST API
+        self.login_as(self.user)
+        response = self.client.get(self.get_url())
+        expected = [{
```
GitEdvard, line 33: The expected outcome from a *-details endpoint should not be a list, it should be a single entry, right?

tests/clims/api/endpoints/test_work_unit.py
```diff
@@ -0,0 +1,104 @@
...
+        url = reverse('clims-api-0-work-units', args=(self.organization.name, ))
+
+        # Wait for getting a large enough response from the endpoint
+        resp = self.wait_for_endpoint_list(url, sample_count)
+        json = resp.json()
+
+        # We expect a list of work units that have only an external ID
+        print(json)
+
+        raise NotImplementedError()
```
GitEdvard: As we agreed, it's ok for you to merge the code as is for now. But I would still like all the tests to pass. I think you agree on that by the way, but I just want to avoid any misunderstandings.

tests/clims/api/endpoints/test_work_unit.py
```diff
@@ -0,0 +1,104 @@
...
+
+        # # Expect a list of WorkUnit items
+        # work_units = workflow.wait_for_work_units("data_entry", sample_count)
+        # print(work_units)
+
+        # raise R
+        # tasks = [t.id for t in tasks]
+
+
+class WorkUnitsByWorkDefinitionEndpointTest(APITestCase):
```
GitEdvard: I think this name is a bit hard to read. I write this comment because I wanted a note on it so that I can find it later...

tests/clims/api/endpoints/test_work_unit.py
```diff
@@ -0,0 +1,104 @@
...
+    def setUp(self):
+        self.clean_workflow_engine_state()
+        self.app.plugins.install_plugins(DemoDnaSeqPlugin)
+        self.has_context()
+
+    @pytest.mark.uc_2_3_2
```
GitEdvard: This is an interesting idea to map use cases to tests!

tests/clims/api/endpoints/test_work_unit.py
```diff
@@ -0,0 +1,104 @@
...
+
+        assert len(data) == count
+        for entry in data:
+            assert len(entry) == 6
+
+            assert entry["id"].startswith("camunda/")
+            assert entry["workflow_provider"] == "camunda"
+            assert len(entry["external_work_unit_id"]) == 36  # expecting guid string
+            assert entry["work_type"] == "clims.plugins.demo.dnaseq.workflows.sequence.DataEntry"
+            assert entry["external_workflow_instance_id"]
```
GitEdvard: There is a missing == here

src/sentry/static/sentry/app/components/sidebar/index.jsx
```diff
@@ -248,15 +248,15 @@ class Sidebar extends React.Component {
                   onClick={this.hidePanel}
                   icon={<icons.AvailableWork />}
                   label={t('Available Work')}
-                  to={`/${organization.slug}/tasks/`}
+                  to={`/${organization.slug}/available-work/`}
```
GitEdvard, line 251: I prefer work-definitions as the endpoint name, so that one knows the name of a single unit. We also have a convention that the endpoints are named in plural, which available-work is not.

GitEdvard: The commit name is confusing ("update task to work unit"). As I understand it, task is the equivalent of work definition, and is not comparable to work unit, which in most cases are "substances".

src/sentry/static/sentry/app/views/availableWork/index.jsx
```diff
@@ -0,0 +1,46 @@
+import ClimsTypes from 'app/climsTypes';
```
GitEdvard, line 1: I have a feeling that this view is replacing another view, which is now redundant. If you don't already have deleted this file, please do it now :)

src/sentry/static/sentry/app/views/availableWork/workDefinitions.jsx
```diff
@@ -2,71 +2,71 @@ import PropTypes from 'prop-types';
...
     }
     return body;
   }
 
-  groupTasksByProcess(taskDefinitions) {
+  groupWorkUnitsByProcess(workDefinitions) {
     const ret = [];
-    const grouped = groupBy(taskDefinitions, (x) => x.processDefinitionKey);
-    for (const [processDefinitionKey, tasks] of Object.entries(grouped)) {
+    const grouped = groupBy(workDefinitions, (x) => x.processDefinitionKey);
```
GitEdvard, line 46: I dont follow here. The method name is "group work unit ...", but here work definitions are grouped.

GitEdvard (commented):
second review:

src/clims/services/workflow/camunda/client.py
withrocks: This is just here because it's still a WIP branch, part of the refactoring I haven't finished yet. I will fix this when I remove the draft label.

I agree about not keeping code around and I only keep it around if it's to be filled in an upcoming pull request. Git is great when it comes to finding old stuff. I recommend checking out `git log -Ssome_function_name` which shows every change to that function name. So when one removes the function, it's easy to find it again with this command. Another option is to just walk through all revisions of the file from the beginning. I do that via extra tooling (vim plugin), but I think pycharm has the same.

src/clims/api/endpoints/work_definition_details.py
withrocks: Yes, I agree.

PR #206, opened 2020-10-26
WIP - CLIMS-421 - transition model

mariya:
This is a WIP because I have some questions that need answering before this is merged. @withrocks please see my comments, feedback is appreciated.


src/clims/models/transition.py
```diff
@@ -0,0 +1,42 @@
...
+    SPAWN = 1
+
+
+class Transition(Model):
+    """
+    Represents a Transition for a Substance.
+    """
+    __core__ = True
+
+    BACKEND_CAMUNDA = 'camunda'
```
mariya: What does this do?

src/clims/models/transition.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # Added because it seems to be used in all CLIMS models.
+    # TODO: check if it is actually needed?
+    organization = FlexibleForeignKey('sentry.Organization')
```
mariya: Will all our CLIMS models link to an Organization object? What is the purpose?

src/clims/models/transition.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # Added in case we want to be able to audit a user's history directly.
+    user = FlexibleForeignKey('sentry.User', related_name='transitions')
+
+    # The work batch in which this transition was created, if it was created in a work batch context.
+    # TODO: should we disallow nulls here? Are transitions ever created outside a work batch context?
+    work_batch = FlexibleForeignKey('WorkBatch', null=True, related_name='transitions')
```
mariya, line 22: Per comment: should we disallow nulls here? Are transitions ever created outside a work batch context?

src/clims/models/transition.py
```diff
@@ -0,0 +1,42 @@
...
+    source_substance = FlexibleForeignKey('Substance', related_name='transitions')
+    target_substance = FlexibleForeignKey('Substance', related_name='transitions')
+
+    transition_type = models.IntegerField(default=TransitionType.MOVE)
+
+    class Meta:
+        app_label = 'clims'
+        db_table = 'clims_transition'
+
+    __repr__ = sane_repr('work_batch_id', 'source_location_id', 'source_substance_id', 'target_location_id', 'target_substance_id')
```
mariya, line 33: How is sane_repr used? It seems that this is a sentry wrapper function for representing an object as a string, but is this representation actually used somewhere?

mariya: Spoke with Edvard now. We're guessing this is used for logging and debugging?

src/clims/models/transition.py
```diff
@@ -0,0 +1,42 @@
...
+    organization = FlexibleForeignKey('sentry.Organization')
+
+    # Added in case we want to be able to audit a user's history directly.
+    user = FlexibleForeignKey('sentry.User', related_name='transitions')
+
+    # The work batch in which this transition was created, if it was created in a work batch context.
+    # TODO: should we disallow nulls here? Are transitions ever created outside a work batch context?
+    work_batch = FlexibleForeignKey('WorkBatch', null=True, related_name='transitions')
+
+    source_location = FlexibleForeignKey('SubstanceLocation', related_name='source_transitions')
```
mariya, line 24: Are reverse accessors here a good idea, or should they be removed?

src/clims/models/transition.py
withrocks: This constant should probably not be here?

src/clims/models/transition.py
withrocks: I think that this doesn't need to be in the model, although a transition is under an organization in the end. The reason is that the transitions always happen in the scope of a `WorkBatch` and they relate to the organization.

src/clims/models/transition.py
```diff
@@ -0,0 +1,42 @@
...
+    target_location = FlexibleForeignKey('SubstanceLocation', related_name='target_transitions')
+    source_substance = FlexibleForeignKey('Substance', related_name='source_transitions')
```
withrocks: I think only the source_location and target_location are really needed here, as they include a link to the substance (and version) in it.

So you can think of `SubstanceLocation` as a pointer to a substance, it knows where it resides.

withrocks:
Looks cool! I have a few comments.

src/clims/models/transition.py
mariya: Removing

src/clims/models/transition.py
mariya: That's useful feedback, thanks. Removing.

PR #207, opened 2020-10-29
clims-438, Implement work endpoint

GitEdvard:
Purpose:
Create endpoint for work-definition-details. This endpoint is used as basis for showing the workbatch-details page in UI, and contains e.g. buttons and textfields to be filled in by user. 

Implementation:
Here, the class workbatchBase is now extensible instead of work-definition, so that work-definition now inherits from workbatchbase instead of extensible. 

Scope of this PR:
This pull reqeust aims at showing the configured buttons in UI. When it comes to trigger a button-click event, it's not finished (names and attributes in endpoint doesn't match UI code). I think this is ok however, since the events endpoint is for now not used anywhere.


src/clims/api/endpoints/events.py
```diff
@@ -6,26 +6,25 @@
 from rest_framework.response import Response
 from rest_framework import status
 from sentry.api.bases.organization import OrganizationEndpoint
-from clims.api.serializers.script_trigger import ScriptTriggerSerializer
+from clims.api.serializers.event import EventSerializer
```
mariya, line 9: Ooh, this is a much nicer name! Great! 👍

src/clims/api/endpoints/work_definition_details.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import absolute_import
```
mariya: Shouldn't this be called WorkBatchDefinitionDetails for consistency?

src/clims/api/serializers/models/work_definition.py
```diff
@@ -0,0 +1,17 @@
+from __future__ import absolute_import
```
mariya: Same question about names: WorkBatchDefinition?

src/clims/configuration/work_definition.py
```diff
@@ -1,9 +1,11 @@
 from __future__ import absolute_import, print_function
 from six import iteritems
+from collections import namedtuple
 from clims.configuration.hooks import HOOK_TAG, HOOK_TYPE
+from clims.services.workbatch import WorkBatchBase
 
 
-class Step:
+class WorkDefinitionBase(WorkBatchBase):
```
mariya: Same question about naming as above.

mariya (approved):
Looks good! I just have some questions/comments about naming, but feel free to merge since issues like this can be addressed later.

GitEdvard:
I agree with that the endpoints should have consistent naming, so I have here changed it in the new commit. However, I'm still concerned about having very long names of domain classes. For example, WorkBatchDefinition is in itself rather long, but how about if you want to compose a descriptive method name with WorkBatchDefinition as the base name to do something with? I'm afraid it will get awkwardly long. I suggest that we can hold this question open, reflect on this and maybe later might reconsider this.

mariya:
Ok! Let's revisit the naming issue when we have time. Merging.

PR #208, opened 2020-11-02
Clims 455 create workbatch example data

mariya:
Added logic to create sample workbatches in `lims createexampledata`. A known issue is that setting the status in the model doesn't seem to write it to the DB, but this is out of scope for this particular ticket. If the reviewer has any idea why this might be happening though, I'd appreciate it!


GitEdvard:
I cannot figure out way status is needed to save workbatch to db. Strange. I merge this!

GitEdvard:
Now, I get why the default initiation doesn't work. The field of status is of kind BoundedPositiveIntegerField, and the default value is zero!

PR #213, opened 2020-11-05
clims-457, implement button clicked event

GitEdvard:
Purpose:
User should be able to trigger an event (ie. run a script) by clicking on a button in UI. 

Implementation:
An event is triggered by specifying the method name that correspond to a specific button, and also specify a workbatch id. This method name is given as the "event" parameter and is fetched when loading the static contents of the page. The workbatch id parameter is needed because a script has to know which set of samples it should act upon. Hence, there is a need for a newly created workbatch when entering the page. In this wip page, it's done in the constructor. Later I think the workbatch should be instantiated before entering the page.

Note: 
The POST api call to the events endpoint still doesn't work, although I have commented out the authentication  lines in the EventEndpoint. Still, I think it would be a good idea to merge this, because it's a lot of name changes and file name changes, which is hell to merge with other PR.


GitEdvard:
I removed the last commit, disabling the authentication on EventEndpoint, since it didn't help (and also made lint-python to fail).

src/clims/configuration/work_batch_definition.py
```diff
@@ -34,34 +32,28 @@ def on_button_click1xx():
...
-        class_dict = dict(cls.__dict__)
-        for _, v in iteritems(class_dict):
-            if cls._matches_event_type(v, event_type):
-                if hasattr(v, HOOK_TAG) and getattr(v, HOOK_TAG) == event_tag:
-                    # TODO: None as self is not pretty
-                    v(None, workbatch)
-                elif not hasattr(v, HOOK_TAG):
-                    v(None, workbatch)
+    def trigger(self, event):
+        class_dict = dict(self.__class__.__dict__)
```
withrocks, line 50: Isn't `__dict__` already a dict? I.e. the `dict` function not necessary?

withrocks:
This looks good to me! I rebased from develop before I merged.

I'm not sure what the issue with the events endpoint post is, because as far as I can see, the test works, which I was sure would catch authentication errors (tests/clims/api/endpoints/test_event.py). If the event was just in the frontend I think I know what it could be - we can discuss that tomorrow. I'm still merging as I agree that we shouldn't wait long with this kind of changes.

src/clims/configuration/work_batch_definition.py
GitEdvard: lol

withrocks: Better to be safe than sorry! :D

PR #215, opened 2020-11-13
clims 448, Show workbatch buttons and fields in UI

GitEdvard:
Purpose:
As the title says. 

Scope:
The button configuration as well as the workbatch properties are now fetched from the backend. Button clicks triggers a matching method call at backend by sending a POST call to the EventEndpoint. Workbatch fields are initialized with db values, and key strokes are reflected on screen. 

Not included in this PR:
Updating workbatch properties by use of a PUT api call. 

Implementation:
When it comes to reflecting key strokes to the screen for property text boxes, a new structure "currentFieldValues" is added to the state. This structure is initialized with the corresponding db values for each field, and will then change to whatever new value the user enters in text boxes.


src/sentry/static/sentry/app/redux/actions/shared.js
```diff
@@ -75,13 +75,25 @@ const acCreate = (resource, urlTemplate) => {
...
+      method: 'POST',
+      data,
+      success: () => {
+        dispatch(acCreateSuccess(resource)(data));
+      },
+      error: (err) => {
+        dispatch(acCreateFailure(resource)(err.statusCode, err.statusText));
+      },
+    });
+    // return axios
```
mariya, line 91: Any reason to keep this commented out code?

src/sentry/static/sentry/app/redux/reducers/index.js
```diff
@@ -19,6 +19,7 @@ const sharedInitialState = {
 
 export const WORK_BATCH_DEFINITION = 'WORK_BATCH_DEFINITION';
 export const WORK_BATCH = 'WORK_BATCH';
+export const WORK_BATCH_DETAILS = 'WORK_BATCH_DETAILS';
```
mariya: These constants and createReducer calls should be in the files specific to the resource (e.g. reducers/workBatch.js). I see that constant definitions started sneaking into the index file before this PR, so maybe we should open a JIRA to clean this up later? The index file should really only have the combineReducers call

src/sentry/static/sentry/app/views/workBatchDetailsWaitingToBeMerged/workbatchDetails.jsx
```diff
@@ -4,102 +4,153 @@ import withOrganization from 'app/utils/withOrganization';
...
+    const arr = Object.values(byIds);
+    let wipWorkbatch = arr[0];
+    return this.props.getWorkBatchDetails(this.props.organization, wipWorkbatch.id);
+  }
+
+  fetchDefinitionsFromProps() {
     let workBatchDefinitionEntry = this.props.workBatchDefinitionEntry;
     if (workBatchDefinitionEntry == null) {
-      return <LoadingIndicator />;
+      return null;
```
mariya: Is there a reason to explicitly return null here, instead of just an empty return statement?

src/sentry/static/sentry/app/views/workBatchDetailsWaitingToBeMerged/workbatchDetails.jsx
```diff
@@ -4,102 +4,153 @@ import withOrganization from 'app/utils/withOrganization';
...
     }
     let {byIds, detailsId} = workBatchDefinitionEntry;
     if (detailsId == null) {
+      return null;
+    }
+    let workDefinition = byIds[detailsId];
+    let {fields, buttons, id} = workDefinition;
+    if (fields === null || buttons === null || id === null) {
```
mariya: This is fine but just FYI, in JS you could just do this as: if(!fields || !buttons || !id) since you can use any falsey type for truth checking

mariya: Unless of course you want to differentiate between null and the empty string, in which case ignore me ;)

src/sentry/static/sentry/app/views/workBatchDetailsWaitingToBeMerged/workbatchDetails.jsx
```diff
@@ -4,102 +4,153 @@ import withOrganization from 'app/utils/withOrganization';
...
 
-  sendButtonClickedEvent(clickedButtonName, workBatchId) {
-    const buttonEvent = {
-      event: clickedButtonName,
-      workBatchId,
-    };
-    this.props.sendButtonClickedEvent(this.props.organization, buttonEvent);
+  initCurrentFieldValues() {
+    let {detailsId, byIds} = this.props.workBatchDetailsEntry;
+    let workbatch = byIds[detailsId];
```
mariya, line 103: Is there some kind of check to ensure this entry exists in ByIds?

src/sentry/static/sentry/app/views/workBatchDetailsWaitingToBeMerged/workbatchDetails.jsx
```diff
@@ -4,102 +4,153 @@ import withOrganization from 'app/utils/withOrganization';
...
+        currentFieldValues,
+      };
+    });
   }
 }
 
 WorkbatchDetails.propTypes = {
   ...ClimsTypes.List,
   organization: ClimsTypes.Organization.isRequired,
+  edvardtext: PropTypes.string,
```
mariya: Should edvardtext be here? :)

mariya (changes requested):
Great work! Just a few comments.

src/sentry/static/sentry/app/redux/actions/shared.js
GitEdvard: Yes. axios calls are currently not working with PUT and POST api calls, because we are missing an authentication header, which either Steinar or me have yet figured out how to write. I have written a ticket for this.

src/sentry/static/sentry/app/redux/reducers/index.js
GitEdvard: Ok

src/sentry/static/sentry/app/views/workBatchDetailsWaitingToBeMerged/workbatchDetails.jsx
GitEdvard: Thanks for the comment, I appreciate that!

src/sentry/static/sentry/app/views/workBatchDetailsWaitingToBeMerged/workbatchDetails.jsx
GitEdvard: Hehe, I actually saw this after I sent in the PR, and wondered if it would be discovered. I'll remove it.

src/sentry/static/sentry/app/views/workBatchDetailsWaitingToBeMerged/workbatchDetails.jsx
GitEdvard: I fix.

src/sentry/static/sentry/app/views/workBatchDetailsWaitingToBeMerged/workbatchDetails.jsx
GitEdvard: I'll add it.

GitEdvard:
Now I have sent in an update according to your comments.

src/sentry/static/sentry/app/redux/actions/shared.js
mariya: Ok! I think it might be a good idea to add a comment with the JIRA ticket number, for context

src/sentry/static/sentry/app/redux/reducers/index.js
mariya: Did you open a ticket or should I?

mariya: Whoops, I missed the latest change.

mariya (approved):
Looks good! Feel free to merge.

PR #216, opened 2020-11-19
Clims 399 transition endpoint

mariya:
Transition service, endpoint, url definition, serializer and model tweaks. There are still some todos as commented in the code and JIRA, but I'd like to get this merged in order to get started on the frontend.


src/clims/api/endpoints/work_batch_transitions.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+
+        # Contract format:
+        # {"source_location_id":1, "target_location": {"container_id": 1, "x":0, "y":0, "z":0}, "transition_type": 1}
+        data = request.data
+
+        # Validate Work Batch
+        # TODO: scope by organization id
+        try:
+            work_batch = WorkBatch.objects.get(pk=int(work_batch_id))
```
GitEdvard: The object used here, work_batch, is a django model. I strongly believe that these should be accessible from the service layer only. So in this case the corresponding workbatch-implementation should be used (inherited from workbatch-base) . See for example substance endpoint.

src/clims/api/endpoints/work_batch_transitions.py
```diff
@@ -0,0 +1,52 @@
...
+        except WorkBatch.DoesNotExist:
+            raise ResourceDoesNotExist
+
+        # Validate Source Location
+        # TODO: Full validation on source and target locations - CLIMS-464
+        try:
+            SubstanceLocation.objects.get(pk=int(data['source_location_id']))
```
GitEdvard: I dont like that "source_location_id" is exposed to the user of the API. According to me, all django internal fields should only be accessible from the service layer. Logic for substance-location is already at place within SubstanceBase, and I think it's preferable to use that class.

src/clims/api/endpoints/work_batch_transitions.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # Currently this validator will always fail since the target location does not yet exist
+        # TODO: Split app.transition.create into two calls,
+        #  the first of which creates the target location record
+        # validator = TransitionSerializer(data=data)
+        # if not validator.is_valid():
+        #    return self.respond(validator.errors, status=400)
+
+        validated = data  # validator.validated_data
+        transition = self.app.transitions.create(validated, work_batch.id)
```
GitEdvard: I would prefer to have a create() method within the serializer (which I suppose will come into this endpoint later on?). Then the code goes like, validator.save(), and within the validator create method, you can call self.app.transitions. I would prefer not to have serializer logic into the transitions service. See "saving instances" in https://www.django-rest-framework.org/api-guide/serializers/

src/clims/services/transition.py
```diff
@@ -0,0 +1,104 @@
...
+            source_location = LocationModel.objects.get(pk=int(source_location_id))
+            assert source_location.current is True
+        except LocationModel.DoesNotExist:
+            raise AssertionError("Source location not found or is not current: '{}'".format(source_location_id))
+
+        substance_id = source_location.substance.id
+        try:
+            substance = self._app.substances.get(id=substance_id)
+        except SubstanceModel.DoesNotExist:
+            raise AssertionError("Source substance not found: '{}'".format(substance_id))
```
GitEdvard, line 37: I would prefer to use the code within Substance base to fetch the location. In that case, the input contract would be something like {source: {substance-id/name, row, col}}.

src/clims/services/transition.py
```diff
@@ -0,0 +1,104 @@
...
+        x = target_location["x"]
+        y = target_location["y"]
+
+        # Z is optional
+        z = 0
+        if "z" in target_location:
+            z = target_location["z"]
+
+        try:
+            container = self._app.containers.get(id=container_id)
```
GitEdvard, line 54: At a transition, a new target container is typically created. Is that done in another preceding call?

tests/clims/services/test_transitions.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # Create container
+        self.register_extensible(MyContainerImplementation)
+        self.container = MyContainerImplementation(name='test-transitions-container')
+        self.container.save()
+
+        # Create samples
+        self.register_extensible(MySampleImplementation)
+        self.samples = []
+        for _ix in range(5):
```
GitEdvard, line 33: I think there is an utility method for creating samples in a container, "create_container_with_samples()"

tests/clims/services/test_transitions.py
```diff
@@ -0,0 +1,128 @@
...
+                y=_ix,
+                z=0,
+                container_version=self.container.version,
+                substance_version=sample.version,
+                current=True)
+            location.save()
+            self.samples.append(sample)
+
+    @classmethod
+    def create_container(cls):
```
GitEdvard, line 50: Here, also, there is the method create_container() in the base test framework.

src/clims/services/transition.py
```diff
@@ -0,0 +1,104 @@
...
+    def __init__(self, app):
+        self._app = app
+
+    # TODO: CLIMS-401 - ensure atomic transaction only commits after plugin logic runs
+    # Atomic transaction:
+    # 1. create target location record
+    # 2. set "current" on source location record to false
+    # 3. create transition record
+    @transaction.atomic
+    def create(self, obj, work_batch_id):
```
GitEdvard: I prefer something more descriptive than "obj"

src/clims/services/transition.py
```diff
@@ -0,0 +1,104 @@
...
+            transition_type=transition_type,
+        )
+        transition.save()
+
+        # TODO: implement MOVE vs SPAWN logic
+
+        return transition
+
+
+class Transition(object):
```
GitEdvard: I have a slight tendency to delay the creation of a class to the moment I actually need it to be used somewhere. The intention of this class would be to provide a history of transitions within a workbatch?  In this case, I can imagine scenarios where this class is being redundant in future, for example if it turns out that there will always only be one transition for each workbatch (as it is in Clarity), the history of the workbatch itself might be sufficient. It might also be a good idea to look for simplifications as this in order to deliver our clims version 1.0 more quickly.

src/clims/services/transition.py
```diff
@@ -0,0 +1,104 @@
...
+    def __init__(self, transition_model):
+        self.archetype = transition_model
+
+    @classmethod
```
GitEdvard: Since this class is kind of empty in behaviour, it is hard for other developers to know the intent of this class.

src/clims/api/endpoints/work_batch_transitions.py
```diff
@@ -0,0 +1,52 @@
...
+            raise ResourceDoesNotExist
+
+        # Currently this validator will always fail since the target location does not yet exist
+        # TODO: Split app.transition.create into two calls,
+        #  the first of which creates the target location record
+        # validator = TransitionSerializer(data=data)
+        # if not validator.is_valid():
+        #    return self.respond(validator.errors, status=400)
+
+        validated = data  # validator.validated_data
```
GitEdvard: is this intentional?

tests/clims/services/test_transitions.py
```diff
@@ -0,0 +1,128 @@
...
+from clims.services.transition import TransitionService
+from clims.services.workbatch import WorkBatchBase
+from clims.services.substance import SubstanceBase
+from clims.services.container import ContainerBase
+from clims.models.transition import TransitionType
+from clims.models.location import SubstanceLocation as LocationModel
+import random
+
+
+class TestTransitionService(TestCase):
```
GitEdvard, line 14: I would like to see a test for the endpoint as well.

GitEdvard (commented):
Great job! I have a few comments. If you agree, I leave it up to you to either merge this now and then send in another PR, or fix it right away. 

In this PR, there are commits that changes the same files and same lines in the files as well. In these situations I tend to merge the commits, because it's easier to read.

mariya:
> Great job! I have a few comments. If you agree, I leave it up to you to either merge this now and then send in another PR, or fix it right away.

I'll fix!
 
> In this PR, there are commits that changes the same files and same lines in the files as well. In these situations I tend to merge the commits, because it's easier to read.

Yeah, agreed. The first two commits were Steinar's, so I don't want to mess with them. The rest of (my) commits I will rebase into a single commit.

src/clims/api/endpoints/work_batch_transitions.py
mariya: Will fix! I'm still learning the architecture here, so I really appreciate feedback like this.

src/clims/services/transition.py
mariya: When a user is at the transition GUI, they would create a transition container first, then start moving/spawning samples to that container, and the transition records simply reference locations in the new target container. A target container can also be a pre-existing one. Please re-open this thread if I have misunderstood something!

src/clims/services/transition.py
mariya: Definitely! I've spent too long away from coding, haha! Fixing.

src/clims/services/transition.py
mariya: Whoops, I think that class was just old boilerplate. I've removed it! (But no, we will never have only one transition per workbatch, unless the workbatch contains only sample? The idea is that a transition is created every time a substance is moved or spawned.)

src/clims/services/transition.py
mariya: Whoops, I think that class was just old boilerplate. I've removed it!

mariya:
I had a long conversation with Steinar and we decided to merge this PR as it is. But all your points will be addressed in a continuation issue! https://snpseq.atlassian.net/browse/CLIMS-470

src/clims/services/transition.py
```diff
@@ -0,0 +1,88 @@
...
+    # Atomic transaction:
+    # 1. create target location record
+    # 2. set "current" on source location record to false
+    # 3. create transition record
+    @transaction.atomic
+    def create(self, transition, work_batch_id):
+        """
+        Creates a new Transition
+        """
+        transition_type = transition["transition_type"]
```
GitEdvard: Previously, I wrote a comment about moving some of the code into the transition serializer. What I foremost reacted on is the use of strings to fetch variables, "transition_type", instead of using an ordinary object as input. I would like to have just that kind of logic focused at the serializer, where it's quite obvious what these strings would be. Here, these key-strings are spread out in a rather large chunk of code. Perhaps you have already addressed this? Then you can of course disregard this comment.

mariya: Duly noted. I've committed some changes that get the worst offenders. The plan is to heavily refactor this later, so I'll let the nested objects slide for now. Thanks for all the feedback, Edvard, it really helped me a lot with understanding the architecture! Going to merge this beast now.

PR #217, opened 2020-11-19
clims-466, Implement update/PUT in shared reduxer

GitEdvard:
Purpose:
Since the update/PUT entry was not yet implemented in the shared action reducers, I implemented it here. In addition, I also fixed so that detailed api error messages are showed in the UI console for the shared action/reducers. 

Method for testing:
For axios calls (not sentry Client calls), I changed the failure actions to receive data from (response.request), instead of just (err). The latter one just didn't return anything. This was discovered by injecting errors in the code, so that the endpoint would reply with an error, trigger that error from the UI, and then inspect the response.

### DIFF
--- a/src/sentry/static/sentry/app/redux/actions/shared.js
+++ b/src/sentry/static/sentry/app/redux/actions/shared.js
@@ -47,9 +47,10 @@ const acGetList = (resource, urlTemplate) => {
     return axios
       .get(url, config)
       .then((res) => dispatch(acGetListSuccess(resource)(res.data, res.headers.link)))
-      .catch((err) =>
-        dispatch(acGetListFailure(resource)(err.statusCode, err.statusText))
-      );
+      .catch((response) => {
+        const message = getErrorMessage(response.request);
+        dispatch(acGetListFailure(resource)(response.request.status, message));
+      });
   };
 };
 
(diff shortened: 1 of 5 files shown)


src/sentry/static/sentry/app/redux/actions/shared.js
```diff
@@ -85,18 +84,64 @@ const acCreate = (resource, urlTemplate) => {
...
+        dispatch(acCreateFailure(resource)(err.status, message));
       },
     });
     // return axios
     //   .post(url, data)
     //   .then(() => dispatch(acCreateSuccess(resource)(data)))
-    //   .catch((err) =>
-    //     dispatch(acCreateFailure(resource)(err.statusCode, err.statusText))
-    //   );
+    //   .catch((response) => {
```
mariya, line 95: Not critical, but I think all these commented out axios calls should have a comment with the JIRA ticket number of the issue where we plan to fix these. Just for context.

src/sentry/static/sentry/app/redux/actions/shared.js
```diff
@@ -85,18 +84,64 @@ const acCreate = (resource, urlTemplate) => {
...
+      data,
+      success: () => {
+        dispatch(acUpdateSuccess(resource)(data));
+      },
+      error: (err) => {
+        const message = getErrorMessage(err);
+        dispatch(acUpdateFailure(resource)(err.status, message));
+      },
+    });
+    // return axios
```
mariya, line 128: Same thing here.

mariya (approved):
This looks great! I have a very minor comment, but it is fine to merge as it is.

PR #219, opened 2020-11-26
clims-442, Update workbatch properties from UI

GitEdvard:
Implement PUT action in workbatch details endpoint, and adjust the serializers for this call. All property fields in UI are here controlled components, with keystrokes captured in a local state. The PUT api call is triggered whenever the user clicks on any user configured button. My first objective here was just to make able to test the update. It's however probable that this call is going to stay there, because the workbatch properties are used within the extensible scripts. At the PUT api call, the local state is transferred to the redux state. 

I made a series of small refactorings in the workbatch details component, according to a PluralSight course I follow.

### DIFF
--- a/src/clims/api/serializers/models/extensible_property.py
+++ b/src/clims/api/serializers/models/extensible_property.py
@@ -5,11 +5,6 @@ from rest_framework import serializers
 
 
 class ExtensiblePropertySerializer(serializers.Serializer):
-    id = serializers.CharField(read_only=True)
-    version = serializers.IntegerField(read_only=True)
     name = serializers.CharField(read_only=True)
     display_name = serializers.CharField(read_only=True)
-    value = serializers.SerializerMethodField(method_name="get_value_with_correct_type")
-
-    def get_value_with_correct_type(self, obj):
-        return obj.value
+    value = serializers.CharField()
--- a/src/clims/api/serializers/models/extensible_property.py
+++ b/src/clims/api/serializers/models/extensible_property.py
@@ -6,5 +6,4 @@ from rest_framework import serializers
 
 class ExtensiblePropertySerializer(serializers.Serializer):
     name = serializers.CharField(read_only=True)
-    display_name = serializers.CharField(read_only=True)
     value = serializers.CharField()
--- a/src/clims/plugins/demo/dnaseq/handlers/create_example_data.py
+++ b/src/clims/plugins/demo/dnaseq/handlers/create_example_data.py
@@ -77,7 +77,7 @@ class DemoCreateExampleDataHandler(CreateExampleDataHandler):
         # Create workbatch to be used at wip workbatch details page. The
         # name of the workbatch has to be known here.
         workbatch = ExampleWorkBatch(name='wip-workbatch', status=1)
-        workbatch.kit_type = "X"
+        # workbatch.kit_type = "X"
         workbatch.save()
         logger.info("Created wip workbatch")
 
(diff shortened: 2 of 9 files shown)


src/clims/plugins/demo/dnaseq/handlers/create_example_data.py
```diff
@@ -77,7 +77,7 @@ def handle(self):
         # Create workbatch to be used at wip workbatch details page. The
         # name of the workbatch has to be known here.
         workbatch = ExampleWorkBatch(name='wip-workbatch', status=1)
-        workbatch.kit_type = "X"
+        # workbatch.kit_type = "X"
```
mariya: Should this be here? (If so please comment why)

src/clims/api/serializers/models/extensible_property.py
```diff
@@ -5,11 +5,5 @@
...
 class ExtensiblePropertySerializer(serializers.Serializer):
-    id = serializers.CharField(read_only=True)
-    version = serializers.IntegerField(read_only=True)
     name = serializers.CharField(read_only=True)
-    display_name = serializers.CharField(read_only=True)
-    value = serializers.SerializerMethodField(method_name="get_value_with_correct_type")
-
-    def get_value_with_correct_type(self, obj):
-        return obj.value
+    value = serializers.CharField()
```
mariya, line 9: Why have all the other fields been removed? (Just for my understanding)

tests/clims/api/endpoints/test_work_batch_details.py
```diff
@@ -32,6 +35,36 @@ def test_get_by_id(self):
         assert response.data['name'] == 'my_workbatch'
         assert response.data['properties']['kit_type']['value'] == 'kit type value'
 
+    @pytest.mark.dev_edvard
```
mariya, line 38: dev_edvard strikes again!

tests/clims/api/serializers/models/test_work_batch_details.py
```diff
@@ -22,6 +25,86 @@ def test_can_serialize_workbatch_details(self):
...
+            'properties': {}
+        }
+        workbatch = MyWorkbatchImplementation(name='my_workbatch')
+        workbatch.save()
+        serializer = WorkBatchDetailsSerializer(workbatch, data=data)
+        assert serializer.is_valid()
+        updated_workbatch = serializer.save()
+        assert updated_workbatch.kit_type is None
+
+    @pytest.mark.dev_edvard
```
mariya, line 92: Another dev_edvard

mariya (changes requested):
Looks great! A couple of very small issues/questions.

mariya:
Feel free to merge!

tests/clims/api/endpoints/test_work_batch_details.py
GitEdvard: Perhaps we could have a discussion about marks in tests. I actually think it's very handy to keep the marks.

src/clims/api/serializers/models/extensible_property.py
GitEdvard: This was something that I had a specific session with Steinar about, asking him if it was okay. The reason was that the deleted fields most likely aren't going to be used in the UI. The extensible property field only return something if it already has a value. At the start, with a blank page, extensible field is an empty array. This means that fields like display-name are fetched from other static calls. Furthermore, "method field" don't allow writes, which we wants here.

src/clims/plugins/demo/dnaseq/handlers/create_example_data.py
GitEdvard: I have now pushed another commit to fix this.

tests/clims/api/endpoints/test_work_batch_details.py
mariya: Sure! Maybe we can use contextual names for the marks though?

PR #220, opened 2020-11-30
clims 471, specific states for "detail" and "list" entries in shared reducer

GitEdvard:
Purpose:
Adjust reducer state in shared reducer so that list reducers, which is connected to endpoints returning lists, have the attribute "byIds" as well as a bunch of other list associated attributes, and "details" reducer, which is connected to endpoints returning single entries, has the "entry" attribute. Because there are now two sets of action/reducer pairs, these are now separated in the "sharedList.js" and "sharedEntry.js" action/reducers.


GitEdvard:
This PR has very little new code in it. Most of the code is moved from one file to another.

withrocks:
I'm looking into conflicts on this one.

withrocks:
I think we should chat a bit about this one before merging so we're on the same page. Taking the others first.

withrocks:
Looks good! It needs to be merged with develop, and I know that the assign samples flow doesn't work fully with this change. Can be merged after that as far as I'm concerned.

PR #221, opened 2020-12-04
CLIMS-470 - Make transition endpoint process transitions in batches, …

mariya:
…allow a more user-friendly API using string indices, and use extensible logic to move/spawn substances instead of accessing models directly


withrocks (commented):
Looks great! I took a quick lock now and will test it next week in relation to a PR I will create.

mariya:
Great, merging!

PR #223, opened 2020-12-09
clims-474 qc analyze script

GitEdvard:
Purpose:
Prepare commonlims framework to be able to handle criteria in workbatches. criteria is added by a technical user at configuration for a step, and is used for validating qc on each sample in workbatch. 

Implementation:
In commonlims-snpseq, analyzing qc criteria could not be imported directly from clarity-ext, since this script is implemented in the lab-logic tool. Instead there is a "native" analyze qc script in snpseq. Because of that, I here added an update queue in workbatch, so that we here can follow the two-step pattern of updating samples, which is 1) put sample in a updated queue when executing script, 2) when script has finished without errors, perform a commit and save to db.

### DIFF
--- a/src/clims/api/serializers/models/work_batch_details_definition.py
+++ b/src/clims/api/serializers/models/work_batch_details_definition.py
@@ -25,3 +25,4 @@ class WorkBatchDetailsDefinitionSerializer(serializers.Serializer):
     name = serializers.CharField(read_only=True)
     buttons = serializers.ListField(child=ButtonField(), read_only=True)
     fields = serializers.ListField(child=FieldPropertyField(), read_only=True)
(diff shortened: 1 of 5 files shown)


withrocks (approved):
Looks good! Merging.

I still think that we should only have criteria in the scripts themselves and not configurable in fields. I think that introduces unnecessary complexity and (if it involves dynamic execution of code, security issues).

In the case of the QC scripts we're looking at now, the REs don't mind if we remove it from the fields and into the scripts only.

An alternative, if the value needs to be set, could be to just have a regular float field named e.g. "max concentration" which the users could then change in the UI. We could also introduce a "range" field (having both min and max), but I think we can go deeper into all of this when have a use case for it.

PR #225, opened 2020-12-15
328 final

withrocks:
A large refactoring where we use the name Work instead of Task. Also includes fixes for enabling assigning samples to a workflow and get into a workbatch view.


src/clims/api/serializers/models/work_batch_details.py
```diff
@@ -8,6 +8,14 @@ class WorkBatchDetailsSerializer(serializers.Serializer):
     id = serializers.IntegerField(read_only=True)
     name = serializers.CharField(read_only=True)
     properties = DictField(child=ExtensiblePropertySerializer(), allow_null=True)
+    created_at = serializers.DateTimeField(read_only=True)
+    updated_at = serializers.DateTimeField(read_only=True)
+
+    # These are still mocked
+    transitions = serializers.SerializerMethodField(read_only=True)
```
withrocks, line 15: All of this stuff I'll revisit next. It was convenient for me to keep it in this PR.

src/clims/migrations/0022_workbatch_input_containers.py
```diff
@@ -0,0 +1,20 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('clims', '0021_auto_20201110_0806'),
+    ]
+
+    operations = [
```
withrocks, line 14: This change wasn't directly changed to this PR, but I'm following up on this in the next PR, when I focus on the work batch itself.

src/clims/services/workflow/camunda/client.py
```diff
@@ -56,79 +57,63 @@ def start_workflow(self, workflow, item):
...
-                               json["name"],
-                               json["formKey"])
+
+            work_unit = WorkUnit(
+                external_work_unit_id=json["id"],
+                workflow_provider=Workflow.BACKEND_CAMUNDA,
+                external_workflow_instance_id=json["processInstanceId"],
+                work_type=json["formKey"])
+            # assert work_unit.work_type, "No work type defined {} {}".format(task_definition_key, process_definition_key)
+            print(json)
```
withrocks: I'll fix this.

src/sentry/static/sentry/app/redux/reducers/workBatchDetails.js
```diff
@@ -12,19 +12,6 @@ const initialState = {
   workBatch: null,
 };
 
-function prepareWorkBatchForLocalWork(workBatch) {
```
withrocks, line 15: This I'll have to revisit in the next PR. I think the component that requires this should do this. But currently this breaks the  transition component (CLIMS-492). But since it was working with mocked data anyways (and I'm the one that's going to fix it) I felt that it was OK.

src/sentry/static/sentry/app/views/availableWorkDetails/index.jsx
```diff
@@ -0,0 +1,146 @@
...
+  }).isRequired,
+};
+
+const mapStateToProps = (state) => {
+  return {
+    workBatch: state.workBatch,
+    workDefinition: state.workDefinition,
+    availableWorkUnit: state.availableWorkUnit,
+
+    // workDefinition:
```
withrocks: I'll fix these comments.

src/clims/migrations/0022_workbatch_input_containers.py
```diff
@@ -0,0 +1,20 @@
...
+        migrations.AddField(
+            model_name='workbatch',
+            name='input_containers',
+            field=models.ManyToManyField(to='clims.Container'),
+        ),
```
GitEdvard, line 19: I guess that container is for the workbatches where work-unit is set to container? For ordinary workbatches, handling samples, input container would be a derived property from the list of substances. I think it might be confusing to both use derived input containers (in ordinary workbatches) and input container meant for workbatches handling containers only.

GitEdvard:
I think that something is wrong here. The last 5 commits seem to be empty.

GitEdvard:
Never mind, now they are back!

src/clims/api/endpoints/work_definition_details.py
```diff
@@ -0,0 +1,38 @@
...
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+
+from sentry.api.base import Endpoint, DEFAULT_AUTHENTICATION
+from clims.utils import single_or_default
+from clims.api.serializers.models.work_definition import WorkDefinitionSerializer
+from clims.api.serializers.models.workunit import WorkUnitSerializer
+
+
+class WorkDefinitionDetailsEndpoint(Endpoint):
```
GitEdvard, line 12: There already exist a class for this that is named work_batch_details_definition.py

withrocks:
I think it might have had to do with rebases.

src/clims/models/workunit.py
```diff
@@ -45,26 +45,82 @@ def tracked_object_global_id(self):
...
 
     class Meta:
         app_label = 'clims'
         db_table = 'clims_workunit'
 
     __repr__ = sane_repr('workflow_provider',
             'external_work_unit_id', 'external_workflow_instance_id')
+
+
+class ExternalWorkUnit(object):
```
GitEdvard, line 108: This sound for me like a work unit that has not yet been saved. Isn't work-unit sufficient for this?

src/clims/services/workflow/camunda/client.py
```diff
@@ -56,79 +57,62 @@ def start_workflow(self, workflow, item):
...
-                               Workflow.BACKEND_CAMUNDA,
-                               None,
-                               json["name"],
-                               json["formKey"])
+
+            work_unit = WorkUnit(
+                external_work_unit_id=json["id"],
+                workflow_provider=Workflow.BACKEND_CAMUNDA,
+                external_workflow_instance_id=json["processInstanceId"],
+                work_type=json["formKey"])
```
GitEdvard, line 103: It should be a serializer doing this

src/clims/api/endpoints/work_batch.py
```diff
@@ -26,13 +28,17 @@ def get(self, request, organization):
...
+        if not validator.is_valid():
+            return Response(validator.errors, status=400)
 
-        logger.debug("Creating workbatches for task list: {}".format(tasks))
+        logger.debug("Creating workbatches for task list: {}".format(validator.validated_data))
+        work_unit_ids = validator.validated_data['work_units']
 
         # Creating a workbatch takes a list of tasks that are ready for a work batch:
-        work_batch = self.app.workflows.create_work_batch(tasks, organization)
+        work_batch = self.app.workflows.start_work(work_unit_ids, organization)
```
GitEdvard, line 41: I'm wondering if it in any cases would be desirable to just create a workbatch and not doing anything else?

src/clims/api/endpoints/work_definition_details.py
```diff
@@ -0,0 +1,38 @@
...
+
+    name = 'clims-api-0-work-definition-details'
+
+    def get(self, request, work_definition_id):
+        # TODO: Here we split the key outside of the service. Move it into it.
+        process_definition_key, work_definition_key = work_definition_id.split(':')
+        work_definitions = self.app.workflows.get_work_definitions(
+            process_definition_key, work_definition_key)
+        ret = single_or_default(work_definitions)
+        return Response(WorkDefinitionSerializer(ret).data)
```
GitEdvard, line 24: It's different serializers for work definition and work definiton details

GitEdvard: I find myself much more comfortable to use the word "step" instead of work definition, just because it's shorter. If we want to be really consistent, it should be "WorkBatchDefinitionSerializer" and "WorkBatchDetailsDefinitionSerializer". Should Details come before Definition? I rather quickly find this very awkward to write, pronounce and reason about.

src/clims/api/endpoints/work_units.py
```diff
@@ -0,0 +1,49 @@
...
+from sentry.api.bases.organization import OrganizationEndpoint
+
+from sentry.api.base import DEFAULT_AUTHENTICATION, Endpoint
+
+from clims.api.serializers.models.workunit import WorkUnitSerializer
+
+logger = logging.getLogger(__name__)
+
+
+class WorkUnitsEndpoint(OrganizationEndpoint):
```
GitEdvard, line 16: Endpoint classes are so far written in singular!

src/clims/api/serializers/models/workunit.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import absolute_import
```
GitEdvard, line 1: There is an inconsistency of the file names, wheather it is an underscore between work_unit or not. I prefer not to have underscore, but we should first of all have consistency here.

tests/clims/api/endpoints/test_work_units.py
```diff
@@ -0,0 +1,58 @@
...
+        cont.save()
+        created_samples_ids = {s.id for s in cont.contents}
+
+        workflow = SequenceSimple()
+        workflow.comment = "Let's sequence some stuff"
+        workflow.assign(cont)
+
+        # 2. Wait for the work to be ready:
+        self.login_as(self.user)
+        work_def_id = "clims.plugins.demo.dnaseq.workflows.sequence.SequenceSimple:DataEntry"
```
GitEdvard, line 37: The last thing, "DataEntry", is a mystery for me. This is no work definition by the way! It's a workflow.

GitEdvard: In other parts of the code, I've refered work defintion by its full class name

src/clims/models/workunit.py
```diff
@@ -45,26 +45,82 @@ def tracked_object_global_id(self):
...
         else:
-            raise AssertionError("Unexpected tracked_object type {}".format(type(value)))
+            raise AssertionError("Unexpected tracked_object model type {}".format(type(model)))
+        self._tracked_object = value
+
+    @property
+    def external_work_unit_key(self):
+        """
+        Returns both the workflow_provider and the external_work_unit_id as one tuple (key).
+        This is required for fully identifying a work unit between different workflow providers.
```
GitEdvard, line 71: As I see it, a new work unit instance is created in the clims db for each external work unit (?). In that case, the inherent id of work unit should suffice to differentiate between work units.

GitEdvard: I'm afraid that there is a risk of fetching of e.g. substance from a workbatch is going to be cumbersome. We have workbatch --> get work units --> get substance. In between the client developer has to dig deep to understand external_work_unit_id, and external_workflow_instance_id. (why use these instead of inherent ids?) My feeling is that this has to be encapsulated in some way.

src/sentry/static/sentry/app/components/workUnit/processListItem.jsx
```diff
@@ -3,35 +3,35 @@ import PropTypes from 'prop-types';
 import React from 'react';
```
GitEdvard, line 3: I cannot interpret what processListItem does based on its name. Process is unclear to me, is it equivalent to a workflow, or is it bound to a workunit and a workflow? process list item doesn't make sense.

src/sentry/static/sentry/app/redux/actions/workDefinition.js
```diff
@@ -0,0 +1,10 @@
+import axios from 'axios';
+import {makeResourceActions} from './shared';
+
+export const RESOURCE_NAME = 'WORK_DEFINITION';
+
+export const workDefinitionActions = makeResourceActions(
+  RESOURCE_NAME,
+  '/api/0/organizations/{org}/NOT-IMPLEMENTED/', // available-work can return work definitions
```
GitEdvard, line 8: As general rule, I think it's best not to add things before they are used anywhere. Going back here in 3 months, it's very hard to know if the code should be saved or is outdated.

src/sentry/static/sentry/app/redux/reducers/availableWorkUnit.js
```diff
@@ -0,0 +1,9 @@
+import {resource} from './shared';
```
GitEdvard, line 1: The name available work unit doesn't make sense. Available work definitions (steps) are the ones with samples queuing to it. As it goes for Clarity, all samples are available to everyone.

src/sentry/static/sentry/app/views/availableWorkDetails/index.jsx
```diff
@@ -0,0 +1,131 @@
+import PropTypes from 'prop-types';
```
GitEdvard, line 1: This must be the view showing when user has clicked on a single workstep to see what samples in it?

GitEdvard (approved):
Nice work! I have some comments, although I think it's more important to get this into the code base, since it's blocking other work.

src/clims/migrations/0022_workbatch_input_containers.py
withrocks: The idea here was that this is the list of containers that need to be shown on the left in the transition component. I think that i was not thinking clearly when I added this! I'd prefer that we just remove it again when we tackle CLIMS-482 (in interest of time).

src/clims/api/endpoints/work_definition_details.py
withrocks: I see :+1:. I threw in a cleanup ticket under the CLIMS-483 epic.

src/clims/models/workunit.py
withrocks: Yeah, or it's one that has been exists in the workflow engine but not in our database.

I used only work unit at first, but since it's not just a regular unsaved Django model, that made it confusing to reason about. I had to have a special field on the model with info that only makes sense when it's a workunit that hasn't been saved yet.

I think actually a great way to resolve both this and the general save issue that you have been talking about (where we are requried to save the django objects in the db in order to save it's relations), is to have higher level business classes for everything, not only substances. They could then take care of all kinds of stuff similar to this. I think we should discuss that in detail when we've implemented the workbatch view. I've added it to CLIMS-505 so we don't forget it.

src/clims/services/workflow/camunda/client.py
withrocks: That's a good point. I hadn't thought of it because I'm interacting with another system.

src/clims/api/endpoints/work_batch.py
withrocks: Right, in this case I just needed a service-style call, one that takes care of a whole business action for me. But it might be needed later.

src/clims/api/endpoints/work_definition_details.py
withrocks: I agree that the naming is not great. I've added an epic for dealing with this at a later point, but we should fix the things that are confusing as soon as we can (although I think we need to merge this one without further change, unless there's something really wrong :)).

I'm also still wondering if we shouldn't call this something else, e.g. Step or UserTask instead. The problem I see with Step, is that we also have non-usertask steps in workflows, so it may be confusing too.

But if this was renamed either, I think we would need:
* [UserTask/Step]DefinitionSerializer (the definition of how the user task should behave, serialized in a list)
* [UserTask/Step]DefinitionDetailsSerializer (the same, serialized in a details view)

Details I think should always be after the Definition word, because it's used to mean that this is a details entry, but not a list entry.

I think we should take this up when we tackle the next task related to this!

src/clims/api/endpoints/work_units.py
withrocks: Right :| I can fix this in a follow-up PR.

src/clims/api/serializers/models/workunit.py
withrocks: I agree.

tests/clims/api/endpoints/test_work_units.py
withrocks: The first part refers to the Workflow class. That's indeed just a regular class. Then that workflow has a "UserTask" in Camunda. It does not necessarily relate to any Python class. Maybe it should, but that would mean that the user needs to write more code. I would prefer that they would only have to write a BPMN diagram and minimum Python code to get going.

withrocks: ... I looked a bit more into this. I think it's how it used to work before and needs to be further fixed. The test doesn't even seem to run here (although everything ran on github). I'll take a peek.

withrocks: OK, I think I'll just fix this in another PR (CLIMS-507, assigned to me). This is how it was coupled up to begin with but it should be the full class name of the work definition, but it is the work definition

src/clims/models/workunit.py
withrocks: This is the key that's required when interfacing with the external workflow engine. The plugin dev will not have to worry about this - but we will :) Sounds good to make it easier to deal with.

src/sentry/static/sentry/app/components/workUnit/processListItem.jsx
withrocks: This is an old name that I forgot to change. I'll throw it in the tech debt pile.

src/sentry/static/sentry/app/redux/actions/workDefinition.js
withrocks: Agreed. The actions is used. It's just the list endpoint that's not used. It's a hack though, and I was confused by this myself. Should just be null.

src/sentry/static/sentry/app/redux/reducers/availableWorkUnit.js
withrocks: The samples may be queued elsewhere, but this the list of work units that are ready for work. But the name is not clear, I agree.

src/sentry/static/sentry/app/views/availableWorkDetails/index.jsx
withrocks: Right. Technically, it will list work units that are in it, but the user sees samples.

withrocks:
Thanks for reading all of this again so thoroughly!

PR #226, opened 2021-01-08
clims 493, Show workbatch details with a non-mock workbatch instance

GitEdvard:
Purpose:
Show workbatch details page with data fetched from db. Here, I have moved the ac create entries in the shared entry reducer to the shared list reducer. I wanted to send in this PR rather quickly, because I suspect this change will affect you other guys in the project.


withrocks (approved):
Looks good!

PR #227, opened 2021-01-15
clims-503, merge work definition endpoints

GitEdvard:
Purpose:
1): Resolve the name clashes in work definition endpoints according to clims-503.
2): Fix a consistent naming of the corresponding action/reducers (shared and work definition)

Implementation:
The name of the action/reducers are updated with a suffix 'Entry' at the end, so that workDefintion.js --> workDefintionEntry.js. I found the former confusing, since it contained a bunch of metadata as well as a workDefinition fetched from db. 

The sharedEntry and sharedList action/reducers contained an attribute 'entry', which I changed to 'resource'. This is how a client call will look like before and after the name change:

Before:
* const workDefinition = this.props.workDefinition.entry

After:
* const workDefinition = this.props.workDefinitionEntry.resource


withrocks (approved):

withrocks:
Looks good to me!

